PR #25, opened 2016-04-24
add missing types int64_t and uint64_t

jameshfisher:


jameshfisher:
Hmm, not sure how these checks passed - I think my PR is wrong, because VeriFast says "Parse error: long long types are not yet supported." I suppose I want `unsigned long int` instead?

jameshfisher:
These old C types are silly. I was actually thinking `unsigned long long int`. But that is not supported by VeriFast either.

Is there any way I can get this to continue? I need to be able to use `uint64_t` one way or another.

btj:
Support for these types was added by 7f1ed34585128f8d68b0f57f87ad74adab12ca5b.

jameshfisher:
Great!

PR #28, opened 2016-04-24
malloc argument should be of type size_t

### LINKED ISSUE #27: malloc argument should not be of type int
I think this is wrong: https://github.com/verifast/verifast/blob/6707c8be1a245a630ecc5c9eafef935ec8ed29c5/bin/malloc.h#L17

[The docs for `malloc`](http://linux.die.net/man/3/malloc) say

```
void *malloc(size_t size);
```

[I'm told that "size_t is an unsigned integer type"](http://stackoverflow.com/questions/2550774/what-is-size-t-in-c).

I can see the `requires` condition states `0 <= size`, but that's different. This should not fail:

```
#include <stddef.h>
#include <stdlib.h>

void foo()
//@ requires true;
//@ ensures true;
{
  size_t size = sizeof(int);
  malloc(size);
}
```

Currently it fails with

```
foo.c(9,10-14): Type mismatch. Actual: uintptr_t. Expected: int.
```

jameshfisher:
Fixes #27

### DIFF
--- a/bin/malloc.h
+++ b/bin/malloc.h
@@ -1,6 +1,8 @@
 #ifndef MALLOC_H
 #define MALLOC_H
 
+#include <stddef.h>
+
 /*@
 
 // In Standard C, freeing a null pointer is allowed and is a no-op.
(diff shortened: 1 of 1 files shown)


btj:
Fixed by 416828da28324fb45b70b6ea75d06abbc0d3e116

PR #44, opened 2016-09-01
Allow unsigned char/short in inductive data types

JoergPfaehler:
Allow unsigned char/short as a parameter to an inductive datatype. Testcase:

/ * @
  inductive test1 = construct_test1(unsigned char);
  inductive test2 = construct_test2(unsigned short);
@ * /

This fixes at least the internal errors of verifast during parsing of this testcase. I am not sure whether the patch will fail at some other place when we actually use the datatype.

### DIFF
--- a/src/verifast1.ml
+++ b/src/verifast1.ml
@@ -1889,7 +1889,7 @@ module VerifyProgram1(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
             let (_, pts) = List.split pts in
             let rec type_is_inhabited tp =
               match tp with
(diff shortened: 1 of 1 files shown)


btj:
Thanks, Jörg! Technically, all integer types are interchangeable within annotations, but it's useful to be able to use the various types to express intent.

JoergPfaehler:
Ah, thanks for the remark! Is the "unsigned" also ignored in annotations, i.e., do they all use the "Int" sort of the SMT solver?

btj:
Yes, it is, and they do.

PR #58, opened 2016-11-14
sizeof(long long) and other

Bromind:
hi, 

Here is something to add support for long long int. 
It adds (updated when I push): 
 - sizeof(long long) which is 8
 - llongs predicate int prelude.h (does the same as ints)

### DIFF
--- a/bin/prelude.h
+++ b/bin/prelude.h
@@ -24,8 +24,8 @@ predicate u_character(unsigned char *p; unsigned char c);
 predicate integer(int *p; int v);
 predicate u_integer(unsigned int *p; unsigned int v);
 
-predicate long_integer(long long *p; long long l);
-predicate u_long_integer(unsigned long long *p; unsigned long long l);
+predicate llong_integer(long long *p; long long l);
+predicate u_llong_integer(unsigned long long *p; unsigned long long l);
 
 predicate short_integer(short *p; short s);
 predicate u_short_integer(unsigned short *p; unsigned short v);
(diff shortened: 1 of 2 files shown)


bin/prelude.h
```diff
@@ -24,6 +24,9 @@ predicate u_character(unsigned char *p; unsigned char c);
 predicate integer(int *p; int v);
 predicate u_integer(unsigned int *p; unsigned int v);
 
+predicate long_integer(long long *p; long long l);
+predicate u_long_integer(unsigned long long *p; unsigned long long l);
```
btj: Wouldn't it make more sense to call these `llong_integer` and `u_llong_integer`?

bin/prelude.h
```diff
@@ -215,6 +218,12 @@ predicate ints(int *p, int count; list<int> vs) =
     :
         integer(p, ?v) &*& ints(p + 1, count - 1, ?vs0) &*& vs == cons(v, vs0);
 
+predicate llongs(long long *p, long long count; list<long long> ls) =
```
btj: Not that it matters (because all integral types mean the same thing inside annotations) but for consistency, it seems better to just use `int` as the type of `count`. (And in any case, the size of an array is orthogonal to the size of its elements.)

bin/prelude.h
Bromind: Yes, absolutely. I did it twice and committed the wrong version... My bad

Bromind:
That should fix the problems you highlighted.

PR #64, opened 2016-11-18
Add cast from (unsigned) long long to floating point types

Bromind:


Bromind:
continuous integration tests fail due to #65.

necto:
Hello, any changes here? @btj, is this PR good enough to merge?

btj:
It's not clear to me that adding a separate conversion function for each integer type is the best approach. An alternative approach would be to have just conversions from `long long` and from `unsigned long long`. But it's not clear that that approach is best either. So I'll go with this approach for now.

PR #86, opened 2017-05-15
Support non-Intel architecture

master-q:
Support non-Intel architecture, such like ARM which has no rdtsc.


master-q:
Thanks! VeriFast runs on Pocket CHIP https://getchip.com/pages/pocketchip!

VeriFast runs on Pocket CHIP

btj:
Wow! That's super-cool! Many thanks! :-)

PR #88, opened 2017-06-01
List available provers in help and in the "No such prover" error message

rafoo:
The help message suggests two values for the `-prover` option (redux and z3) but calling verifast with `-prover z3` fails when verifast is compiled with support for z3v4.5.

This commit adds a listing of supported provers both in the output of `verifast -help` and in the error message reported when a bad prover name is given to the `-prover` option.


btj:
Thanks, Raphaël!

PR #97, opened 2017-09-14
Allow "typedef struct s typ" without actual declaration of s

keigoi:
C specification (and gcc at least) seems to allow ``typedef struct`` without actual declaration of that struct, while Verifast rejects it.

I.e., C allows the code below

```c
typedef struct s typ;
```
without actual declaration of ``s``. And this patch lets Verifast accept it while bulding ``typedefmap``. 

I believe this does not compromise soundness; if such a structure is referred in the actual code,  Verifast simply raises an exception ``Not_found`` from somewhere. 

From that reason, such structures should also be added to ``structmap``, however, I currently couldn't find the right place to do so.

### DIFF
--- a/src/verifast1.ml
+++ b/src/verifast1.ml
@@ -1374,7 +1374,7 @@ module VerifyProgram1(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
   
   (* Region: check_pure_type: checks validity of type expressions *)
   
-  let check_pure_type_core typedefmap1 (pn,ilist) tpenv te =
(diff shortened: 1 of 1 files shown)


keigoi:
I think such ``typedef``s are very common in actual C projects; for example, we are currently trying to parse (a part of) BIND9 and found a lot of them in a header file. An instance can be found at:

https://source.isc.org/cgi-bin/gitweb.cgi?p=bind9.git;a=blob;f=lib/isc/include/isc/types.h

which has a bunch of ``typedef``s without actual declaration of the ``struct``s (which are in other headers and not necessarily included from some .c files).

keigoi:
Hooray!

PR #104, opened 2017-12-12
Add an `assert_term` method to the prover API

rafoo:
The `assert_term` method is identical to `assume` but without checking for
unsatisfiability afterward.

In the case of the Z3 backends, this PR removes all ignored calls
to `(check-sat)` and might thus improve performance but I have not
benchmarked it.


src/redux.ml
```diff
@@ -918,6 +918,10 @@ and context () =
       else
         assumes_with_more_pending_splits <- assumes_with_more_pending_splits + 1;
     
+    method assert_term t =
+      self#register_pending_splits_count;
+      ignore (self#assume_internal t)
+
```
btj: Please copy all statistics-gathering and verbosity code from method assume (but with "Redux.assume" -> "Redux.assert_term")

btj:
Looks great, thanks! I just have the one comment on the Redux patch.

Also, would you mind if I rebase these commits, or would that cause pain?

rafoo:
I just modified `assert_term` in redux as you requisted. I have nothing against a rebase.

PR #118, opened 2018-01-25
Introduce predicate_nonempty

let-def:
This implements the `predicate_nonempty` construct which was discussed on the mailing-list.
It asserts that a predicate maintains some parts of the heap alive, making it easier to pass the recursion termination check.

Note:
- the patch rewrites a lot of parts because it adds one field to the predicate tuple definition, so all deconstructions of the tuple are touched... would you accept a patch rewriting this tuple as a record?
- I remove trailing spaces from files before committing, however there are a lot of trailing ones in verifast codebase, would you accept a patch that remove all trailing spaces (not doing any other change)?

### DIFF
--- a/bin/prelude.h
+++ b/bin/prelude.h
@@ -18,19 +18,19 @@ lemma void div_rem(int D, int d);
     requires d != 0;
     ensures D == D / d * d + D % d &*& abs(D % d) < abs(d) &*& abs(D / d * d) <= abs(D);
 
-predicate character(char *p; char c);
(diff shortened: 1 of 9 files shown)


btj:
Many thanks, Frédéric, for your excellent work, and apologies for the late reply.

Yes, both of the cleanups you suggest would be very welcome. I guess you want to do them before I merge this PR?

btj:
Abandoned PR? Closing for now.

PR #120, opened 2018-01-28
Allow hardtab in string literals

master-q:
Allow hard tab in string literals.

"6.4.5 String literals" at C99 puts following

```
Syntax
  string-literal:
    " s-char-sequence opt "
    L" s-char-sequence opt "
  s-char-sequence:
    s-char
    s-char-sequence s-char
  s-char:
    any member of the source character set except
      the double-quote ", backslash \, or new-line character
    escape-sequence
```

Above means that we should only maintain new-line character in string literals at lexer.ml.

### DIFF
--- a/src/lexer.ml
+++ b/src/lexer.ml
@@ -662,7 +662,7 @@ let make_lexer_core keywords ghostKeywords startpos text reportRange inComment i
             Stream.Failure -> error "Bad string literal."
         in
         store c; string ()
-    | c when c < ' ' -> raise Stream.Failure
+    | c when c = '\n' -> raise Stream.Failure
     | c -> text_junk (); store c; string ()
   and char () =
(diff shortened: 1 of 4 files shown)


src/lexer.ml
```diff
@@ -662,7 +662,7 @@ let make_lexer_core keywords ghostKeywords startpos text reportRange inComment i
             Stream.Failure -> error "Bad string literal."
         in
         store c; string ()
-    | c when c < ' ' -> raise Stream.Failure
```
btj: Does this do the right thing when EOF occurs inside a string literal? EOF is represented in the lexer as a '\000' character.

btj:
Abandoned PR? Closing for now; to reopen, please reply to my review.

PR #128, opened 2018-03-24
Support z3 with opam

master-q:
You will see following error, if you use opam to install z3.

```
$ make
--snip--
  OCAMLOPT  z3v4dot5prover.cmx
cp: cannot stat '/home/kiwamu/.opam/system/lib/z3/../../libz3.so': No such file or directory
make: *** [GNUmakefile:159: ../bin/libz3.so] Error 1
```

But the `libz3.so` file is located at following:

```
$ ocamlfind query -qe Z3 || ocamlfind query z3
/home/kiwamu/.opam/system/lib/z3
$ ls -l /home/kiwamu/.opam/system/lib/z3/libz3.so
-rwxr-xr-x 1 kiwamu kiwamu 23283608 Jan 19 11:58 /home/kiwamu/.opam/system/lib/z3/libz3.so*
```

### DIFF
--- a/src/GNUmakefile
+++ b/src/GNUmakefile
@@ -100,8 +100,15 @@ ifndef NOZ3
 endif
 Z3V4DOT5PRESENT=false
 ifdef Z3V4DOT5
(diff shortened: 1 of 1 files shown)


btj:
Closing for now; to reopen, please provide a patch that does not break the CI builds.

PR #162, opened 2019-03-18
Use ocamlfind for ocaml-num

bno1:
I couldn't compile verifast on Arch Linux using system packages because on Arch the num package is installed in `/usr/lib/ocaml/num` instead of `/usr/lib/ocaml`. With this change `ocamlfind` is used to locate the proper path for num.

~I also removed `nums.cmxa` from the vfide rules because it isn't needed.~


btj:
Thanks!

PR #175, opened 2019-06-16
[Tool] Lemma Post-Condition Explorer

RamirezLucas:
I created a command-line tool that allows to search for patterns in the post-condition of lemmas. The objective is to make the process of writing proofs faster by enabling users to easily figure out if certain things have already been proven. It takes the form of a separate executable that is built by doing `make` in the `src/` directory. Once built, the tool is located in `bin/` under the name `explorer`. Run `./explorer --help` for detailed usage instructions.

The user must specify in CLA's a list of directories containing VeriFast-annotated files (`*.h`, `*.c`, or `*.gh`) from which the tool builds a repository of lemmas that will then be searched for user-provided patterns. By default the tool runs in interactive-mode and prompts the user for patterns indefinitely, displaying the list of matching lemmas each time.

A few important things:

* patterns must follow standard VeriFast syntax
* `_` are used as placeholders just like in VeriFast, and will match with anything
* variable names that start with a `_` (e.g. `_xs`) in a pattern are considered as named placeholders that can represent any VeriFast expression, while others are considered as real variables in the AST sense. Some examples:
    1. pattern `map(fst, _xs)` will match with `map(fst, drop(n,ys))` but not with `map(snd, drop(n,ys))`
    2. pattern `map(_fst, _xs)` will match with `map(fst, drop(n,ys))` and with `map(snd, drop(n,ys))`
    3. pattern `take(_, _xs) == _xs` will match with `take(n, ys) == ys` but not with `take(n, ys) == nil`

I hope this tool can help VeriFast users in their work, and am open to suggestions to improve the program.


btj:
Many thanks for this, Lucas!! Sounds very useful; I'll be happy to merge it in once it builds.

The Travis build failures are probably caused by an out-of-date `.depend` file; you may need to rebuild it using `make depend`.

RamirezLucas:
Thank you for your feedback! I am glad you like the idea!
`make depend` solved the issue indeed.

PR #199, opened 2020-05-23
Add termination checking support for direct recursion

jpdoyle:
This patch adds support for recursive calls by rewriting the precondition of recursive calls as `[1]call_perm_(currentThread,foo) &*& (P)`.

It seems to work (and doesn't seem to allow arbitrary recursion), but I'm not entirely confident in this change. I'm especially suspicious of the scope parameters of `WVar`, but `LocalVar` and `FuncName` seem like they match?

### DIFF
--- a/examples/termination/ackermann.c
+++ b/examples/termination/ackermann.c
@@ -19,20 +19,11 @@ fixpoint nat A(nat m, nat n) {
 
 @*/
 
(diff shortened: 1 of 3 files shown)


jpdoyle:
Ah sorry, messed up my git branching. Reopening!

PR #204, opened 2020-06-26
vfide: user can set tab width in preferences

nchong-at-aws:


btj:
Thanks, Nathan!

BTW: `vfide` also supports specifying the tab size on the first line of the file, like so:
```c
// tab_size:4
...
```
The `tab_size:N` token can be anywhere on the line.

Also: to avoid having to set your preference again at each run, you may also want to add a `vfide` command-line option, which you can then set in your custom launch script for `vfide`.

nchong-at-aws:
Thanks Bart! I will have a go with the comment-directive. And that's a good idea about the command-line option: I'll make that change.

Issue #206, opened 2020-07-02
`old_*` doesn't seem to work with 3 levels of tuerk-style loops


jpdoyle:
This example changes its verification status when the `#if 0` is toggled:
```c
int foo()
    /*@ requires true; @*/
    /*@ ensures  result == 3*42; @*/
    /*@ terminates; @*/
{
    int x = 0;
    int i,j,k;

    for(i = 0; i < 6; ++i)
        /*@ requires i >= 0 &*& i <= 6 &*& x + 3*7*(6-i) < 100000; @*/
        /*@ ensures  old_x + (6-old_i)*7*3 == x; @*/
        /*@ decreases 6-i; @*/
    {
        for(j = 0; j < 7; ++j)
            /*@ requires j >= 0 &*& j <= 7 &*& x + 3*(7-j) < 100000; @*/
            /*@ ensures  old_x + (7-old_j)*3 == x; @*/
            /*@ decreases 7-j; @*/
        {
#if 0
            x += 3;
#else
            for(k = 0; k < 3; ++k)
                /*@ requires k >= 0 &*& k <= 3 &*& x + (3-k) < 100000; @*/
                /*@ ensures  old_x + (3-old_k) == x; @*/
                /*@ decreases 3-k; @*/
            {
                ++x;
                /*@ recursive_call(); @*/
            }
#endif
            /*@ recursive_call(); @*/
            /*@ {
                assert old_x + 3 + (7-(old_j+1))*3 == x; 
                assert old_x + (7-old_j)*3 == x; 
            } @*/
        }
        /*@ recursive_call(); @*/
    }
    return x;
}
```
With only 2 nested loops it seems to behave correctly.

jpdoyle:
This example changes its verification when the `#if 0` is toggled:
```c
int foo()
    /*@ requires true; @*/
    /*@ ensures  result == 3*42; @*/
    /*@ terminates; @*/
{
    int x = 0;
    int i,j,k;

    for(i = 0; i < 6; ++i)
        /*@ requires i >= 0 &*& i <= 6 &*& x + 3*7*(6-i) < 100000; @*/
        /*@ ensures  old_x + (6-old_i)*7*3 == x; @*/
        /*@ decreases 6-i; @*/
    {
        for(j = 0; j < 7; ++j)
            /*@ requires j >= 0 &*& j <= 7 &*& x + 3*(7-j) < 100000; @*/
            /*@ ensures  old_x + (7-old_j)*3 == x; @*/
[...]

PR #207, opened 2020-07-05
Prioritize the old value of `x` over the current value of `old_x`.

jpdoyle:
Fixes https://github.com/verifast/verifast/issues/206


btj:
Excellent job, Joe! Many thanks!!

PR #212, opened 2020-08-11
Support sizeof(bool) and bools_inv

SolalPirelli:
This is necessary and sufficient to make the following program verify:

```c
#include <stdbool.h>
#include <stdlib.h>

void x(bool* bs)
//@ requires bools(bs, ?c, ?lst);
//@ ensures bools(bs, c, lst);
{
        //@ assert length(lst) == c;
}

int main(int argc, char** argv)
//@ requires true;
//@ ensures true;
{
        bool* bs = malloc(10 * sizeof(bool));
        if (bs == 0) return 1;
        //@ assert malloc_block_bools(bs, 10);
        //@ assert bools(bs, 10, ?bs_list);
        //@ assert length(bs_list) == 10;
        x(bs);
        free(bs);
        return 0;
}
```

### DIFF
--- a/bin/crt.vfmanifest
+++ b/bin/crt.vfmanifest
@@ -71,6 +71,7 @@
 .provides ./prelude.h#ullongs_inv
 .provides ./prelude.h#shorts_inv
 .provides ./prelude.h#ushorts_inv
+.provides ./prelude.h#bools_inv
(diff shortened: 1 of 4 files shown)


SolalPirelli:
Oops, I should've looked at the meaning of k more carefully... thanks 😄

PR #214, opened 2020-08-17
Add '#pragma once' support

SolalPirelli:
This avoids the need to wrap each header in ifdef/define/endif.

I added support only at the beginning of a file; technically, `#pragma once` can appear anywhere in the file with the same effect (GCC and Clang both support this), but this is unlikely to happen in real code and I didn't see an obvious way to handle it in VeriFast's preprocessor, so I added an error in that case instead.


btj:
Looks great! Many thanks!

PR #225, opened 2021-01-22
pretty print json

NielsMommen:
Pretty printing for json values, which can be useful when debugging


btj:
Note that you can also use `python -m json.tool` to pretty-print JSON:

```
verifast -c -json myfile.c | python -m json.tool
```

PR #232, opened 2021-04-02
Improve sizeof handling, including packed structures

SolalPirelli:
I hope you're not getting tired of my pull requests 😬 

I have code that uses struct arrays and wanted to be able to reason about converting an array of chars to an array of structs, which requires knowing what size the structs are (or adding runtime checks, but that's annoying), and that can only be done if they're packed since otherwise the compiler is free to insert padding (and thus one cannot even reason about >1 struct since e.g. `2 * sizeof(struct xyz)` could overflow).

This PR does the following:
- Add an assumption that `sizeof` returns a value `<= SIZE_MAX`
- Add support for `__attribute__((packed))` (see `examples/sizeof/packed.c`)
- Fix the `usbkbd` example, as it was declaring members of types that were only defined later (which is illegal in C)
- Add an assumption that `float` and `double` are 4 and 8 bytes respectively, i.e., IEEE-754.

That last point is maybe a bit controversial, but I think assuming IEEE-754 is reasonable? Otherwise I could change the code to only care about field sizes if the struct is packed, and just not support floats/doubles in packed structs.

By the way, the `usbkbd` example cannot compile with GCC due to the presence of nested comments, which VeriFast handles fine but are not in the C spec. Is there some other compiler that can handle these, or is this example a VeriFast-only thing?

### DIFF
--- a/examples/sizeof/packed.c
+++ b/examples/sizeof/packed.c
@@ -1,0 +1,31 @@
+#include <stdint.h>
+#include <stdlib.h>
+
+struct x {
+	uint64_t a;
+	uint8_t b;
+} __attribute__((packed));
+
+struct y {
+	struct x value;
+};
+
+void foo(void)
+//@ requires emp;
+//@ ensures emp;
+{
+	// Proper size
+	//@ assert sizeof(struct x) == 9;
+
+	// No offset inside
+	struct x value;
+	//@ assert (char*) &(value.b) == (char*) &value + sizeof(uint64_t);
+
+	// No padding
+	struct y* allocated = malloc(sizeof(struct y));
+	if (allocated != NULL) {
+		//@ leak malloc_block_y(allocated);
+		//@ leak x_a(&(allocated->value), _);
+		//@ leak x_b(&(allocated->value), _);
+	}
+}
--- a/examples/sizeof/struct.c
+++ b/examples/sizeof/struct.c
@@ -1,0 +1,17 @@
+#include <stdint.h>
+
+struct x {
+	uint64_t a;
+	uint64_t b;
+};
+
+void foo(void)
+//@ requires emp;
+//@ ensures emp;
+{
+	//@ assert sizeof(struct x) >= 0;
+	//@ assert sizeof(struct x) <= SIZE_MAX;
+
+	// The compiler can add alignment or padding, so this may not be true
+	//@ assert sizeof(struct x) == 16; //~ should_fail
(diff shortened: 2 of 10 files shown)


btj:
Many thanks!! Do keep these PRs coming!

As for the nested comments in `usbkbd`: yes, the file being compiled is not the file that was verified, unfortunately. I don't remember if/how we checked that they matched.

Yes, assuming IEEE-754 seems fine to me, too. In fact, I wonder if it would not also be fine to make some assumptions about the layout of non-packed structs. After all, do the ABIs of the various platforms not fix the struct layout for those platforms? For example: the [System V Application Binary Interface AMD64 Architecture Processor Supplement Draft Version 0.99.6](https://refspecs.linuxfoundation.org/elf/x86_64-abi-0.99.pdf) on p. 12 specifies the alignments and sizes of the scalar types, and on p. 13 says the following:
> Structures and unions assume the alignment of their most strictly aligned component. Each member is assigned to the lowest available offset with the appropriate alignment. The size of any object is always a multiple of the object‘s alignment.
> An array uses the same alignment as its elements, except that a local or global array variable of length at least 16 bytes or a C99 variable-length array variable always has alignment of at least 16 bytes.
> Structure and union objects can require padding to meet size and alignment constraints. The contents of any padding is undefined.

The [Procedure Call Standard for the Arm® 64-bit Architecture (AArch64)](https://github.com/ARM-software/abi-aa/blob/master/aapcs64/aapcs64.rst#composite-types) also specifies struct layout, and it seems to match the System-V ABI pretty exactly.

See also the [x64 software conventions](https://docs.microsoft.com/en-us/cpp/build/x64-software-conventions?view=msvc-160#types-and-storage) document, part of the Microsoft Visual C++ docs.

Would it not be very safe to assume that the alignment of any scalar type is at most 32 bytes, and therefore the size of a struct is at most the sum of the sizes of its members, each rounded up to a multiple of 32? This rough upper bound should eliminate the need for a run-time arithmetic overflow check when computing the size of an array of structs in many cases.

SolalPirelli:
Indeed, a 32-byte upper bound on alignment seems safe and useful for code that hardcode sizes to small constants. Then one could add some predicate like `structure(void*, size_t, ???)` where `???` is a representation of the structure, and then add support for `calloc` by hardcoding the case when it's passed a `sizeof` expression just like `malloc`... though I have no idea what that `???` should be to be useful. :-)

PR #233, opened 2021-04-04
Make travis build scripts run on MacOS Mojave

markrtuttle:
Minor changes needed to make the ./setup-build.sh and ./build.sh scripts complete on MacOS 10.14 Mojave.

The setup-build.sh script was testing`[ $TRAVIS = "true" ]` which was failing with a syntax error as `[ = "true"]` when TRAVIS was not set.

The GNUmakefile was using `!` to invert the return code from grep, but the grep was finding the requested lines and returning 0, and the inversion was causing the build to fail.

### DIFF
--- a/setup-build.sh
+++ b/setup-build.sh
@@ -25,7 +25,7 @@ elif [ $(uname -s) = "Darwin" ]; then
       brew update
   fi
 
-  if [ $TRAVIS = "true" ]; then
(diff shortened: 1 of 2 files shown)


btj:
The grep issue was fixed in 9526146d95477d72c3d463cc1926336c7af1fcba .

PR #236, opened 2021-05-27
C++ support

NielsMommen:
Adds basic C++ support to verify simple C++ programs.


btj:
How hard would it be to support Windows as well?

NielsMommen:
> How hard would it be to support Windows as well?
Dune is used to build the OCaml dependencies that are required for C++ support. Currently there's a problem on Windows related to the bulid directory. Dune does not recognize the cydrive directory and creates a new one at another location. Therefore, no vfdeps package exists for Windows. I haven't found a solution to the problem yet.

btj (commented):
I finally had a look---sorry for the wait! Cool stuff!!! Before looking further at the internals of what you've done, I have some questions about `src/ast.ml`.

btj:
BTW: I did a build on my machine (using `brew install llvm@11`); works like a charm! I checked the release package (GitHub Actions artifact) as well; it just works too! Great stuff!!

btj:
> > How hard would it be to support Windows as well?
> 
> Dune is used to build the OCaml dependencies that are required for C++ support. Currently there's a problem on Windows related to the bulid directory. Dune does not recognize the cydrive directory and creates a new one at another location. Therefore, no vfdeps package exists for Windows. I haven't found a solution to the problem yet.

I'm having a go at adding the extra dependencies to https://github.com/verifast/vfdeps-win-noz3 . I've been able to build dune-based dependencies just fine. One problem I had was that when building dune-configurator, dune.exe was trying to overwrite itself, which produces a "Permission denied" on Windows. The solution was to simply use the dune.exe from the source directory rather than the installed one. See my progress at https://github.com/verifast/vfdeps-win-noz3/tree/cxxdeps .

btj:
https://github.com/verifast/vfdeps-win-noz3/tree/cxxdeps now builds fine with all of the extra dependencies for C++ support. I haven't tried to build VeriFast against it yet, though.

btj:
Next problem: the LLVM binary release for Windows does not include the `LLVMConfig.cmake` or `llvm-config.cmake` files, so it looks like we need to build LLVM from source (sigh...).

btj:
For building LLVM/Clang on Windows, the instructions provided are for building with Visual Studio. If we build LLVM using Visual Studio, the question is if we can then still build our libtooling client using the Cygwin-to-MINGW cross-compiler, which is the only Windows toolchain supported by OCaml. That is, can the Cygwin-to-MINGW toolchain consume the libtooling library (a C++ library) produced by Visual Studio? Seems doubtful.

NielsMommen:
> Next problem: the LLVM binary release for Windows does not include the `LLVMConfig.cmake` or `llvm-config.cmake` files, so it looks like we need to build LLVM from source (sigh...).

Does this relate to the binary that is downloaded through the GitHub Action or from somewhere else?

src/cxx_frontend/libtooling/VerifastASTExporter.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+  int err = tool.run(&factory);
+
+  capnp::MallocMessageBuilder result;
+  auto serResult = result.initRoot<stubs::SerResult>();
+  if (err)
+    serResult.setErr();
+  else
+    serResult.setOk();
+  capnp::writeMessageToFd(1, result);
```
btj, line 101: On Windows, use `_setmode(1, _O_BINARY);` to disable LF-to-CRLF conversion on stdout.

btj:
OK, I managed to get `testsuite.cxx.mysh` to pass on Windows.

Starting from a fresh Windows Server 2019 Datacenter VM on Google Compute Engine (an `e2-standard-4` machine with 100GB SSD storage, accessed from my Macbook using Microsoft's Remote Desktop app installed from the App Store; you need to set the Windows password for the instance in the Google Cloud Console; this delivers a fairly comfortable development experience at some 0.25 EUR per hour; make sure to shut the machine down when you're not using it and to delete the disk eventually), I did the following:

1. Install Chrome by opening a PowerShell terminal and entering
      ```
      $LocalTempDir = $env:TEMP; $ChromeInstaller = "ChromeInstaller.exe"; (new-object    System.Net.WebClient).DownloadFile('http://dl.google.com/chrome/install/375.126/chrome_installer.exe', "$LocalTempDir\$ChromeInstaller"); & "$LocalTempDir\$ChromeInstaller" /silent /install; $Process2Monitor =  "ChromeInstaller"; Do { $ProcessesFound = Get-Process | ?{$Process2Monitor -contains $_.Name} | Select-Object -ExpandProperty Name; If ($ProcessesFound) { "Still running: $($ProcessesFound -join ', ')" | Write-Host; Start-Sleep -Seconds 2 } else { rm "$LocalTempDir\$ChromeInstaller" -ErrorAction SilentlyContinue -Verbose } } Until (!$ProcessesFound)
      ```
1. Install Git for Windows from git-scm.com. In the installer, make sure to choose NO CRLF conversion.
2. Install Python from python.org. Choose to register Python in the PATH in the installer.
3. Install Visual Studio 2019 Community. In the installer, choose Desktop development using C++. You only need VC++, the Windows 10 SDK, and the CMake tools. (You do not need to install CMake separately.)
4. Download the LLVM sources: `git clone --depth 1 --branch llvmorg-11.1.0 https://github.com/llvm/llvm-project` (This is MUCH faster than downloading the source zip and extracting it using Windows Explorer.)
5. In an x86 Native Tools Command Prompt, go to `C:\llvm-project` and do
      ```
      mkdir build
      cd build
      cmake -DLLVM_ENABLE_PROJECTS=clang -DLLVM_TARGETS_TO_BUILD=X86 -G "Visual Studio 16 2019" -A Win32 -Thost=x64 ..\llvm
      start llvm.sln
      ```
      The latter opens the Solution in Visual Studio. Build the *Clang libraries* project. Go do something else; this takes about an hour or so. (There should be a way to do this on the command line using MSBuild but I didn't manage to figure it out yet.)
6. Again in an x86 Native Tools Command Prompt, do
      ```
      cd \
      git clone --depth 1 --branch v0.9.0 https://github.com/capnproto/capnproto
      cd capnproto/c++
      mkdir build
      cd build
      cmake -G "Visual Studio 16 2019" -A Win32 -Thost=x64 ..
      msbuild -p:Configuration=Debug -p:Platform=Win32
      ```
7. Again in an x86 Native Tools Command Prompt, do
      ```
      cd \
      git clone --branch cxx https://github.com/NielsMommen/verifast
      cd verifast\src\cxx_frontend\libtooling\build
      notepad ..\CMakeLists.txt # Change the script to find the Cap'n Proto libraries in `C:\capnproto\c++\build\src\capnp\Debug`, the KJ libraries in `C:\capnproto\c++\build\src\kj\Debug`, and the include files in `C:\capnproto\c++\src`
      cmake -DLLVM_INSTALL_DIR=C:/llvm-project/build -DCAPNP_CXX_DIR=C:/capnproto/c++ -A Win32 -Thost=x64 ..
      msbuild VF_AST_Exporter.sln -p:Configuration=Debug -p:Platform=Win32 -A Win32 -Thost=x64 ..
      ```
    The `vf-cxx-ast-exporter.exe` executable is now in `C:\verifast\src\cxx_frontend\libtooling\build\Debug`.
9. Set up the VeriFast build environment using `setup-windows.sh`, but use https://github.com/verifast/vfdeps-win-noz3/releases/tag/0.1 .
10. Adapt the VeriFast makefiles to not skip building the C++ frontend on Windows.
11. In a Cygwin prompt, run
      ```
      export PATH=C:\vfdeps\bin:$PATH
      export GTK=C:\vfdeps
      cd /cygdrive/c/verifast/src
      make
      ```

PR #253, opened 2021-11-19
Fix dynamic linking to redundant libraries

### LINKED ISSUE #246: Verifying C++ code fails due to missing LLVM dependencies
The VeriFast nightly release for macOS fails with
```
Fatal error: exception (Failure
   "Cxx AST exporter error:\
  \ndyld: Library not loaded: /usr/local/opt/llvm/lib/libLLVM.dylib\
  \n  Referenced from: /Users/bartj/verifast/verifast-21.04-20-g797e8fcf/bin/vf-cxx-ast-exporter\
  \n  Reason: image not found\
  \n")
```
Please test in a clean environment (e.g. run `sudo mv /usr/local/Cellar /usr/local/_Cellar` and similar for other dirs, or run in a fresh VM).

The release for Windows also fails, with
```
VeriFast internal error:
(Failure
  "the Cxx frontend received a serialized result which was undefined.")
```

NielsMommen:
Fixes #246.

`LLVM/Clang` binaries from package managers or GitHub releases may dynamically link to `LLVMlib` and redundant libraries that are not needed in `VeriFast`. 

- We now use a `LLVM/Clang`[release](https://github.com/NielsMommen/vf-llvm-clang-build/releases/tag/v1.0.0)  with a configuration that meets `VeriFast`'s requirements and does not dynamically link to redundant libraries.
- On Windows, `vcruntime140.dll` and `msvcp140.dll` are shipped with `VeriFast`.
- The custom `LLVM/Clang` build also allows to simplify the CMake script for the `vf-cxx-ast-exporter`: one configuration is used for every platform.


config.sh
```diff
@@ -1 +1,2 @@
 export VFDEPS_NAME=vfdeps-7c315b0
+export VF_LLVM_CLANG_BUILD_VERSION=a5d9179
```
btj, line 2: This commit is no longer in any branch of the repo? Did you force-push?

NielsMommen: Indeed. I didn't realise that I was working on the repo's main branch and it contained a lot of WIP commits. I squashed and forced pushed to clean it. Do you want me to rerun the workflow so the artifact's version points to the latest commit?

btj: Yes, please. (And while you're at it, perhaps re-introduce a dash between `vf-llvm-clang-build` and the version?) And check that the tag associated with the GitHub release corresponds to the commit that was built.

btj:
Good job, Niels! Many thanks!

PR #260, opened 2022-01-27
Stack (de)allocation

NielsMommen:
- Constructor contract is checked when an object is allocated on the stack
- Destructor contract is checked when an object is deallocated on the stack (or goes out of scope)

### DIFF
--- a/src/verifast.ml
+++ b/src/verifast.ml
@@ -666,13 +666,20 @@ module VerifyProgram(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
           match globals with
             [] -> cont h
           | (x, (_, tp, addr, init))::globals ->
-            let init =
-              match !init with
-                None -> Default
-              | Some e -> Expr e
-            in
-            produce_c_object l coef addr tp eval_h init true true h env $. fun h env ->
-            iter h globals
+            begin match language, dialect with 
+            | CLang, Some Cxx ->
+              let verify_ctor_call = verify_ctor_call (pn, ilist) leminfo funcmap predinstmap sizemap tenv ghostenv h env addr (Some x) in 
+              produce_cxx_object l real_unit addr tp eval_h verify_ctor_call !init true h env @@ fun h _ ->
+              iter h globals
+            | _ ->
+              let init =
+                match !init with
+                  None -> Default
+                | Some e -> Expr e
+              in
+              produce_c_object l coef addr tp eval_h init true true h env $. fun h env ->
+              iter h globals
+            end
         in
         iter h globalmap
       end $. fun h ->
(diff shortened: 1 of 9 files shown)


btj (commented):
Very nice! But are all of these case analyses between C and C++ necessary? It seems better to avoid them if possible. Is the behavior of a C program different when interpreted as a C++ program when it comes to stack allocation/deallocation?

NielsMommen:
> Very nice! But are all of these case analyses between C and C++ necessary? It seems better to avoid them if possible. Is the behavior of a C program different when interpreted as a C++ program when it comes to stack allocation/deallocation?

The amount of match cases has been reduced. A language check is still needed when destroying an object. The `init_expr` is checked for the creation of objects in order to determine if the constructor contract has to be checked. `produce_cxx_object` and `consume_cxx_object` now delegate control to `produce_c_object` and `consume_c_object` respectively when required.

src/verifast.ml
```diff
@@ -669,19 +726,21 @@ module VerifyProgram(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
...
-            produce_c_object l real_unit addr t eval_h init true true h env $. fun h env ->
+            let verify_ctor_call = verify_ctor_call (pn, ilist) leminfo funcmap predinstmap sizemap tenv ghostenv h env addr (Some addr_name) in
+            produce_cxx_object l real_unit addr t eval_h verify_ctor_call init true true h env @@ fun h env ->
             iter h ((x, envTp)::tenv) ghostenv ((x, addr)::env) xs
           in
           match t with
             StaticArrayType (elemTp, elemCount) ->
             produce_object t
-          | StructType sn when !address_taken || (let (_, fds, _, _) = List.assoc sn structmap in match fds with Some fds -> List.exists (fun (_, (_, gh, _, _, _)) -> gh = Ast.Ghost) fds | _ -> true) ->
+          | StructType sn when !address_taken || (language = CLang && dialect = Some Cxx) || (let (_, fds, _, _) = List.assoc sn structmap in match fds with Some fds -> List.exists (fun (_, (_, gh, _, _, _)) -> gh = Ast.Ghost) fds | _ -> true) ->
```
btj, line 743: It would be better if we could support structs-as-values in C++ as well. Instead of checking for C++, check for the presence of non-C members (i.e. constructors, destructors, or methods)?

btj: See https://github.com/verifast/verifast/commit/bf7dc7f662a4c6340cf0fdf9e211334265216e1d

btj: The main use case would be passing structs by value as function arguments or return values.

btj (commented):
Looks good! (The PR is still marked as Draft?)

src/verifast.ml
btj: In general, it would be good if there were no "regressions": every C program would continue to verify when interpreted as C++.

NielsMommen: > It would be better if we could support structs-as-values in C++ as well. Instead of checking for C++, check for the presence of non-C members (i.e. constructors, destructors, or methods)?

Currently, when no constructor/destructor is found, an error is given that they should be defined explicitly (because we do not generate default constructors/destructors yet). Checking for the presence of a constructor would also be hard because its mangled name would be required to query its existence.

We cannot check for the existence for methods, because they are translated to functions, i.e., there is no way to easily check if a function would be a member function of an object.

Another possibility would be to add a flag in structs/unions. Clang allows to check if a record is a [POD type](https://en.cppreference.com/w/cpp/named_req/PODType).  That would however probably bloat the code.

NielsMommen:
> Looks good! (The PR is still marked as Draft?)

Should be unmarked now.

src/verifast.ml
btj: OK; let's leave it like this for now, then.

PR #261, opened 2022-01-28
Add index_of_update lemma

dc-mak:


btj:
You have to implement the lemma in `bin\list.c`. But the way you stated it, it doesn't hold. `x` might appear at an index less than `j`.

jpdoyle:
A counterexample:
```c
/*@ #include <list.gh> @*/

/*@

lemma void index_of_update<t>(t x, int j, t y, list<t> xs);
  requires 0 <= j && j < length(xs);
  ensures index_of(x, update(j, y, xs)) == (x == y ? j : index_of(x, xs));

lemma void unsound()
    requires true;
    ensures  false;
{
    index_of_update(2,2,2,{1,2,3});
}

@*/
```

PR #264, opened 2022-03-11
Fix unmatched case when referencing a struct name in ghost code in C

NielsMommen:


src/verifast1.ml
```diff
@@ -1644,7 +1644,7 @@ module VerifyProgram1(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
         Some (ld, _, _) when dialect = Some Cxx -> 
         reportUseSite DeclKind_Struct ld l;
         StructType id
-      | None ->
+      | _ ->
```
btj, line 1647: There's another case of this bug a few lines below...

PR #269, opened 2022-04-19
C++ single inheritance

NielsMommen:
Simple single inheritance in C++ without virtual bases and methods.


src/verifast1.ml
```diff
@@ -1885,8 +1897,16 @@ module VerifyProgram1(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
...
-        let rec iter1 fmap fds has_ghost_fields =
+        let lte, bases_size = 
+          match bases with
+          | [] -> (* no bases, a struct cannot have size 0 -> lt *)
+                  ctxt#mk_lt, int_zero_term
+          | _ ->  (* only match on non-virtual bases because virtual bases are not supported yet -> throws unmatched case otherwise *)
+                  (* compiler may introduce padding in bases and the derived struct's members may fit in this padding -> le *)
+                  ctxt#mk_le, bases |> List.fold_left (fun size (CxxBaseSpec (l, base, false)) -> struct_size_partial smapwith0 l base |> ctxt#mk_add size) int_zero_term
+        in
+        lte bases_size s |> ctxt#assert_term;
```
btj, line 1908: Are you sure? If the compiler can put fields of the derived struct inside the padding of base structs, can it not also put one base inside the padding of another base? For example, consider
```c++
struct A { int x; short y; } // 2 bytes of padding at the end
struct B { char c; }
struct C : A, B {}
```
Could it be that `sizeof(struct A) == 8` and `sizeof(struct C) == 8`?

Bottom line: how useful is it to assert lower bounds on the size of a derived struct? What use cases do you envisage  that would need this information? Perhaps do not bother for now?

tests/cxx/single_inheritance.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+/*@
+predicate B_pred(B *b; int a_i, int b_i) =
+	A_pred(b, a_i) &*& b->i |-> b_i;
+@*/
+
+struct B : A {
+    int i;
+
+    B() : i(1)
```
btj, line 39: Also test the case where the derived constructor explicitly calls the base constructor with some (nonzero number of) arguments?

btj: Also test the case where a derived constructor delegates to some other derived constructor which then calls the base constructor with some arguments (that use the derived constructor's parameters)?

btj (changes requested):

src/verifast1.ml
NielsMommen: Removed the assertions

src/verifast1.ml
```diff
@@ -655,8 +655,13 @@ module VerifyProgram1(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
       * asn (* post *)
       * bool (* terminates *)
       * (stmt list * loc) option option
+    type base_spec_info = 
+        loc 
+      * string
+      * bool (* virtual *)
     type struct_info =
         loc
+      * base_spec list
```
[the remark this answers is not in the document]
NielsMommen: `(base_spec_info map * struct_field_info map) option` is now the optional body of a struct

tests/cxx/single_inheritance.cpp
NielsMommen: Added delegating calls

PR #270, opened 2022-05-20
C++ multiple inheritance

NielsMommen:
Multiple inheritance without virtual bases or methods

### DIFF
--- a/src/verifast1.ml
+++ b/src/verifast1.ml
@@ -658,6 +658,7 @@ module VerifyProgram1(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
     type base_spec_info = 
         loc 
       * bool (* virtual *)
+      * termnode (* offset *)
     type struct_info =
         loc
       * (base_spec_info map * struct_field_info map) option (* None if struct without body *)
@@ -1216,7 +1217,6 @@ module VerifyProgram1(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
       | Struct (l, sn, body_opt, attrs)::ds ->
         begin match body_opt with
         | Some (bases, _) ->
-          if List.length bases > 1 then static_error l "Multiple inheritance is not supported." None;
           if List.exists (fun (CxxBaseSpec (l, _, is_virtual)) -> is_virtual) bases then static_error l "Virtual inheritance is not supported." None;
         | _ -> ()
         end;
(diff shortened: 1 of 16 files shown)


src/verifast1.ml
```diff
@@ -2520,13 +2522,43 @@ module VerifyProgram1(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
...
+    | Some (_, Some (bases, _), _, _) -> check_bases bases 
+    | None -> false
   
   let rec compatible_pointees t t0 =
     match (t, t0) with
       (_, Void) -> true
     | (Void, _) -> true
     | (PtrType t, PtrType t0) -> compatible_pointees t t0
-    | StructType x, StructType y when dialect = Some Cxx && is_derived_struct_of x y -> true
+    | StructType x, StructType y when dialect = Some Cxx && is_derived_of_base x y -> true
```
btj, line 2561: This seems wrong. Actually, the old version seems wrong as well, but this seems even more wrong. The old version was already wrong because `compatible_pointees` and `unify` are supposed to be symmetric: `unify t1 t2` should hold if and only if `unify t2 t1` holds. Another reason why the old version was wrong is that it is dangerous to allow an implicit conversion from `Derived**` to `Base**`, because this suggests incorrectly that the pointer points to a variable into which any value of type `Base*` can be stored. Technically speaking: pointers are invariant, no covariant (and not contravariant).

The new version is even more wrong because implicitly converting a `Derived**` to a `Base**` where the offset of `Base` in `Derived` is nonzero suggests incorrectly that dereferencing that pointer will yield a valid pointer to an object of type `Base`.

Implicit widening conversions, like from `Derived` to `Base`, should be enabled in `expect_type_core`, not in `unify`.

(Not that any of this matters for soundness; in VeriFast, all pointer types really mean the same thing. But if we allow this, we might as well allow implicit conversion from any `Foo*` to any `Bar*`.)

src/verifast1.ml
```diff
@@ -2548,7 +2580,7 @@ module VerifyProgram1(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
       for_all2 unify ts1 ts2 && inputParamCount1 = inputParamCount2 && inductiveness1 = inductiveness2
     | (ArrayType t1, ArrayType t2) -> unify t1 t2
     | (PtrType t1, PtrType t2) -> compatible_pointees t1 t2
-    | StructType x, StructType y when dialect = Some Cxx && is_derived_struct_of x y -> true
+    | StructType x, StructType y when dialect = Some Cxx && is_derived_of_base x y -> true
```
btj, line 2583: Dus ook fout.

btj (changes requested):
Looks good! Just a few tweaks needed.

NielsMommen:
Looks like a part of the last commit message got dropped because of backticks. Should have been: "C++: check upcast in `expect_type_core`"

PR #283, opened 2022-06-07
Check for null terminator in lemma string_to_chars

eleehiga:
To the string_to_chars lemma in the post condition, added the assertion of the string having a null terminator at the end. This will satisfy the precondition of the null terminator check for chars_to_string. This addition helps when doing a string_to_chars to something and then later having a chars_to_string to convert it back.

### DIFF
--- a/bin/prelude.h
+++ b/bin/prelude.h
@@ -543,7 +543,7 @@ lemma_auto void chars_to_string(char *s);
 
 lemma_auto void string_to_chars(char *s);
     requires [?f]string(s, ?cs);
-    ensures [f]chars(s, length(cs) + 1, append(cs, cons('\0', nil))) &*& !mem('\0', cs);
+    ensures [f]chars(s, length(cs) + 1, append(cs, cons('\0', nil))) &*& !mem('\0', cs) &*& index_of('\0', append(cs, cons(0, nil))) == length(cs);
 
 lemma_auto void chars_separate_string(char *s);
     requires [?f]chars(s, ?n, ?cs) &*& mem('\0', cs) == true;


jpdoyle:
This can be proven without changing the postcondition, by calling
https://github.com/verifast/verifast/blob/3a752a7c07967c4ec8042e746cee4a56dcf849ba/bin/listex.gh#L56

eleehiga:
@jpdoyle But with this update to the post condition it is easier to call `string_to_chars` and soon after `chars_to_string`. Perhaps it is more work to call `string_to_chars`, then the `lemma_auto(index_of` lemma, and then the `chars_to_string` lemma.

jpdoyle:
If you include `listex.gh` and add an autolemma for `take(length(a),append(a,b)) == a`, you don't need any extra proof:

```c
/*@ #include <listex.gh> @*/

/*@

lemma_auto(take(length(a),append(a,b)))
void take_append_exact<t>(list<t> a, list<t> b)
    requires true;
    ensures  take(length(a),append(a,b)) == a;
{ switch(a) { case nil: case cons(x,xs): take_append_exact(xs,b); } }

lemma void test_lemma(char* p)
    requires [?f]string(p,?s);
    ensures  [ f]string(p, s);
{
    string_to_chars(p);
    chars_to_string(p);
}

@*/
```

eleehiga:
@jpdoyle For a new user like me, it would be hard to come to the solution of putting that `lemma_auto` to solve the error that appears when doing `string_to_chars(p); chars_to_string(p);`. Therefore then you guys could possibly update the tutorial with what is a `lemma_auto` or perhaps accept this change I proposed.

jpdoyle:
You can also use `string_to_body_chars` and `body_chars_to_string` instead, which don't require you to prove things about `index_of` and other such list functions:
```c
/*@
lemma void test_lemma(char* p)
    requires [?f]string(p,?s);
    ensures  [ f]string(p, s);
{
    string_to_body_chars(p);
    body_chars_to_string(p);
}
  @*/

PR #301, opened 2022-07-08
Verify that all bases are constructed when a method is called

NielsMommen:
[C++ draft n4910](https://github.com/cplusplus/draft/releases/tag/n4910) §11.9.3 - Initializing bases and members [class.base.init]/16:

> Member functions (including virtual member functions, 11.7.3) can be called for an object under construction. (...) However, if these operations are performed in a ctor-initializer (or in a function called directly or indirectly from a ctor-initializer) before all the mem-initializers for base classes have completed, the program has undefined behavior.


btj:
Nice job!

Issue #303, opened 2022-08-02
aarch64: embedded use of "x86intrin.h" and "__rdtsc" in "src/linux/caml_stopwatch.c"


stephengaito:
"Fools rush in where angels fear to tread" :-(

I am attempting to get verifast running on a raspberry pi 4 8GB (aarch64)....

In the first instance I compiled directly on my raspberry pi 4 inside an ubuntu:focal (20.04) container using podman  (took a while to do llvm-clang.... but it is now done).

When I finally got to compile verifast itself, I encountered the fact that the `src/linux/caml_stopwatch.c` code uses the `x86intrin.h` header and more specifically the gcc `__rdtsc` builtin function to directly access the x86 time stamp counter....

(I *think* that this is the "only" direct x86-ism in the verifast code -- certainly everything else "compiles" ... )

I have found a couple of discussions on how to replace `__rdtsc` on aarch64 ([Fun with Timers and cpuid](https://cpufun.substack.com/p/fun-with-timers-and-cpuid) and [AArch64 Programmer's Guides: Generic Timer](https://developer.arm.com/-/media/Arm%20Developer%20Community/PDF/Learn%20the%20Architecture/Generic%20Timer.pdf?revision=c710e7a7-9f52-4901-8c9d-91b19f44f9c7)) so I suspect I *could* hack this bit of code to work....

Before I do this... **I have a question: what is the stopwatch "for"? How critical is the stopwatch to verifast? How sensitive is the stopwatch to the "exact" rate of "ticks"?** (note that the "Fun with timers" article suggests that `__rdtsc` is *not* very accurate).

stephengaito:
If it helps anyone, the attached patch seems to fix this problem well enough to allow the system to compile ...

[caml_stopwatch.c.patch.txt](https://github.com/verifast/verifast/files/9241404/caml_stopwatch.c.patch.txt)

btj:
The stopwatch code is irrelevant to VeriFast's correct operation; it is used only to provide the timings in the `-verbose` listings.

I created https://github.com/verifast/verifast/pull/306 from your patch. If you want to be properly recorded as the author, please create one yourself.

PR #306, opened 2022-08-04
Make stopwatch code compile on Arm64

btj:
Patch provided by stephengaito

### DIFF
--- a/src/linux/caml_stopwatch.c
+++ b/src/linux/caml_stopwatch.c
@@ -1,9 +1,26 @@
 #define _GNU_SOURCE
 #include <sched.h>
-#include <x86intrin.h>
+
+// see: https://stackoverflow.com/a/69856463
+#ifdef __x86_64__
+  #include <x86intrin.h>
+#endif
+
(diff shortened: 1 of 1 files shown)


stephengaito:
@btj 
While being added as an author would be flattering, it is not critical. ( #303 )

However, I do not suggest accepting this patch *until* we have worked out how to regularly build *for* aarch64 using your github workflow.... 

SO I suspect I should "step up to the plate" and add the workflow rules (in a fork) and then contribute both my patch and the aarch64 workflow as a pull request back to the main repository.

Alas, as I have never done a cross-compile before.... so... it might take me a couple of days... (or longer) to get to this step...

btj:
VeriFast now has a MacOS-AArch64 build https://github.com/verifast/verifast/pull/666 . I had forgotten about this PR and your proposed code; I used `clock_gettime_nsec_np(CLOCK_UPTIME_RAW)` instead, as recommended by Apple [here](https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time).

PR #325, opened 2022-10-25
[java] Add extra inheritance check

NielsMommen:


examples/java/override_methods/NoOverrideTargetParam.java
```diff
@@ -0,0 +1,30 @@
...
+        return i;
+    }
+}
+
+
+class B extends A //~
+{
+    int j;
+    
+    //@ predicate valid() = this.valid(A.class)() &*& this.j |-> ?m_j;
```
btj, line 20: Define `B.valid` in such a way that it does *not* imply `A.valid` (and define `A.valid` as `false` and write `A.getI()` so that it crashes)! For this example as it is now, the fact that VeriFast does not accept it is unfortunate; show an example where it is essential that VeriFast does not accept it.

src/verify_expr.ml
```diff
@@ -1067,7 +1067,9 @@ module VerifyExpr(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
         let overrides =
           flatmap
             begin fun (sign, MethodInfo (lm, gh, rt, xmap, pre, pre_tenv, post, epost, pre_dyn, post_dyn, epost_dyn, terminates, ss, fb, v, is_override, abstract, mtparams)) ->
-              if is_override || pre != pre_dyn || post != post_dyn then [(cn, sign)] else []
+              if is_override || pre != pre_dyn || post != post_dyn then
+                 [(cn, sign)] 
+              else []
```
btj, line 1072: Spurious diff; please revert

btj (changes requested):

PR #330, opened 2022-11-07
[C++] Instance predicates

NielsMommen:


src/parser.ml
```diff
@@ -1706,9 +1741,10 @@ and
 
 end
 
-let parse_decls lang dataModel enforceAnnotations ?inGhostHeader =
+let parse_decls (lang, dialect_opt) dataModel enforceAnnotations ?inGhostHeader =
```
btj, line 1744: Is it really necessary to pass a dialect argument to the parser? Are there any programs that are valid in two languages but have a different meaning? It is okay if the VeriFast parser accepts more programs than the compiler. Also, note that we can consider some words as keywords in one language, and regular identifiers in another language.

src/verifast1.ml
```diff
@@ -397,6 +397,7 @@ module VerifyProgram1(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
   let ancestry_symbol = mk_symbol "ancestry" [ctxt#type_int] ctxt#type_inductive Uninterp
   let ancester_at_symbol = mk_symbol "ancester_at" [ctxt#type_int; ctxt#type_int] ctxt#type_int Uninterp
   let get_class_symbol = mk_symbol "getClass" [ctxt#type_int] ctxt#type_int Uninterp
+  let cxx_typeid_symbol = mk_symbol "typeid" [ctxt#type_int] ctxt#type_int Uninterp
```
btj, line 400: Since the provenance commits, the provertype of a pointer is now `inductive`, not `int`.

src/verifast1.ml
```diff
@@ -1823,6 +1838,18 @@ module VerifyProgram1(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
   let classterms = classterms1 @ classterms0
   let interfaceterms = interfaceterms1 @ interfaceterms0
 
+  let cxx_type_info_map1 =
+    structdeclmap |> List.fold_left begin fun acc (sn, (_, body_opt, _)) ->
+      match body_opt with
+      | None ->
+        acc
+      | Some _ ->
+        let type_info = get_unique_var_symb (sn ^ "_type_info") cxx_type_info_ref_type in
```
btj, line 1847: The provertype of `RefType` is currently still `int`; it should be `inductive` to match the provertype of pointers (since references are really just pointers).

btj: Try running the C++ examples with Z3; it does typechecking of the terms.

src/verifast1.ml
```diff
@@ -3099,6 +3130,65 @@ module VerifyProgram1(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
...
+        in
+        match preds_in_bases bases_map preds_in_struct with
+        | [] -> 
+          sn, get_unique_var_symb (sn ^ "#" ^ pred_name) (PredType ([], [], None, Inductiveness_Inductive))
+        | [(pred_fam, param_map0, pred_symb)] ->
+          if not (for_all2 (fun (x, t) (x0, t0) -> expect_type_core pred_loc "Predicate parameter covariance check" (Some true) t t0; true) param_map param_map0) then
+            static_error pred_loc "Predicate override check: parameter count mismatch." None;
+          pred_fam, pred_symb
+        | _ ->
+          static_error pred_loc "Ambiguous override: multiple overriddeb predicates." None
```
btj, line 3178: Typo

src/ast.ml
```diff
@@ -288,6 +288,9 @@ and
   | RealLit of loc * num * float_literal_suffix option
   | StringLit of loc * string (* string literal *)
   | ClassLit of loc * string (* class literal in java *)
+  | CxxTypeid of loc * expr (* typeid(expr) in C++ *)
+  | WCxxTypeid of loc * expr * type_ (* typechecked typeid(expr) *)
+  | CxxTypeInfo of loc * type_ (* TypeInfo object in C++ *)
```
btj, line 293: As mentioned, I will be using the type info stuff for C programs as well, to check effective types.

src/verifast.ml
```diff
@@ -2897,6 +2910,7 @@ module VerifyProgram(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
       execute_branch @@ fun () ->
       with_context (Executing ([], env, loc, sprintf "Verifying constructor '%s'" struct_name)) @@ fun () ->
       assume_neq this_term (null_pointer_term ()) @@ fun () ->
+      assume_eq (ctxt#mk_app cxx_typeid_symbol [this_term]) (List.assoc struct_name cxx_type_info_map) @@ fun () ->
```
btj, line 2913: Could this not be a base ctor call? I.e. the most derived type of `this_term` might be a subtype of `struct_name`?

src/verifast.ml
```diff
@@ -2970,6 +2984,7 @@ module VerifyProgram(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
       execute_branch @@ fun () ->
       with_context (Executing ([], env, loc, sprintf "Verifying destructor '%s'" @@ cxx_dtor_name struct_name)) @@ fun () ->
       assume_neq this_term (null_pointer_term ()) @@ fun () ->
+      assume_eq (ctxt#mk_app cxx_typeid_symbol [this_term]) (List.assoc struct_name cxx_type_info_map) @@ fun () ->
```
btj, line 2987: Same question

src/verifast.ml
```diff
@@ -2835,11 +2835,24 @@ module VerifyProgram(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
...
+    in
     let init_fields delegated h struct_addr this_type init_list env ghostenv leminfo sizemap current_thread cont =
       (* when a delegated ctor was called, no initializeation is needed for the fields *)
       if delegated then cont h else
+      begin fun cont ->
+        assume_field_ptr_limits struct_addr cont
+      end @@ fun () ->
       let init_field h field_name field_type value struct_addr cont =
-        assume_field h struct_name field_name field_type Real struct_addr value real_unit cont
+        assume_field h struct_name field_name field_type Real struct_addr value real_unit false cont
```
btj, line 2855: The extra boolean argument is to avoid re-assuming the same fact again? The marginal performance benefit seems less than the code complexity cost...

btj:
Otherwise, looks great! Good job!

src/verifast.ml
NielsMommen: §11.9.5 Construction and destruction [class.cdtor]/5: `(...) When typeid is used in a constructor (including the mem-initializer or default member initializer (11.4) for a non-static data member) or in a destructor, or used in a function called (directly or indirectly) from a constructor or destructor, if the operand of typeid refers to the object under construction or destruction, typeid yields the std::type_info object representing the constructor or destructor’s class. (...)
`

src/parser.ml
NielsMommen: Removed the optional dialect

src/verifast1.ml
NielsMommen: Changed the domain and range to `inductive` type

src/verifast1.ml
NielsMommen: Changed the provertype of `RefType` to `inductive`

src/ast.ml
NielsMommen: Removed the `Cxx` prefix from `CxxTypeInfo`

src/verifast.ml
btj: Interesting! But this means that the type id of a given pointer changes during the lifetime of that pointer? So we cannot model the typeid as a pure function of the pointer? (Otherwise, I think you can prove false by stating typeid(this) == typeid(struct A) in the postcondition of the constructor for struct A; this will yield a contradiction when constructing a subclass object.

btj: But do we need `typeid(E)` yet in this PR? After all, there are no virtual methods yet, so there is no need yet for dynamic binding of instance predicates? Perhaps we should deal with expression typeids and with virtual methods in the same PR (because I suspect similar mechanisms will be necessary), and in this PR we do not yet support typeid(E).

NielsMommen: Makes sense. I'll remove support in this PR for `typeid(E)` where `E` is not a type expression.

btj: In fact, does it make sense to have virtual instance predicates (= predicate families indexed by a typeid) in a non-virtual class (= a class without a vtable)? Would it not be cleaner and more C++-like to have non-virtual instance predicates by default and have virtual instance predicates only if marked by keyword `virtual`?

btj: In Java, instance predicates are always virtual because methods are always virtual.

btj: At this point, I think it makes most sense to put this PR on hold and do the virtual member functions PR first.

NielsMommen: Do you also need the functionality to evaluate a `typeid(struct S)` expression to a `S_type_info` term?

btj: Yes, I do. Thanks!

tests/cxx/instance_preds/Shape.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+  Square(const Pos &pos, int width) : Shape(pos), m_width(width)
+  //@ requires pos.valid(?x, ?y) &*& width >= 0;
+  //@ ensures valid(x, y) &*& pos.valid(x, y) &*& Square_vtype(this, thisType);
+  {
+    //@ close valid(x, y);
+  }
+
+  ~Square()
+  //@ requires valid(_, _) &*& Square_vtype(this, thisType);
```
btj, line 141: I guess this destructor overrides the Shape destructor. Do you check spec compatibility? This succeeds here because `Square_vtype` is auto-opened to `Shape_vtype`? But what if Square has another polymorphic superclass as well? At first sight, it seems that virtual destruction of objects with multiple polymorphic superclasses is tricky because the `vtype` chunk of the other base class is not available? But we should postpone this issue until later if possible.

btj (commented):
This review is WIP but I already want to comment on the issue of polymorphic destruction. We should postpone this issue if possible, but I just want to signal it as a point of attention for the future.

tests/cxx/instance_preds/Shape.cpp
NielsMommen: Indeed, virtual destruction of objects with multiple polymorphic base classes would not succeed because the virtual destructor's spec in the derived class is checked against all virtual base destructor specs. It is (currently) also not supported to override multiple instance predicates from base classes in a derived class.

tests/cxx/instance_preds/Shape.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+int calcArea(const Shape &s)
+//@ requires s.valid(?x, ?y);
+//@ ensures s.valid(x, y);
+{
+  return s.calcArea();
+}
+
+void destroy(Shape *s)
+//@ requires s->valid(_, _) &*& new_block_Shape(s);
```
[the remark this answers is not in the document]
NielsMommen: I did not adjust `new` and `delete`. New'ing a Square produces a `new_block_Square`. `delete p` calls `p`'s destructor and consumes its `P_new_block` afterwards. Calling a virtual destructor on `p` first consumes and reproduces P_vtype(p, ?p_type) to retrieve `p`'s `vtype`. Then, the destructor is called with `thisType = p_type`. This means that `destroy` in the example can only be used on an object with most derived type `Shape`.

Does it make sense to remove the prefix from `prefix_new_block` and have `new_block(void *p, std::type_info *p_type` instead? New'ing `S` would then call the constructor of `S` and produce `new_block(s, &typeid(S))`. `delete s` then first consumes `new_block(s, _)` and afterwards calls the destructor. I think this would make deleting objects with multiple inheritance work too.

Explicitly calling a destructor only makes sense when placement new (which we don' support yet) was used. We currently do not support calling a destructor explicitly (I should make the error message in that case more clear).

tests/cxx/instance_preds/Shape.cpp
btj: 👍 Fine like this for now.

tests/cxx/instance_preds/Shape.cpp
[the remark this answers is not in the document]
btj: > New'ing a Square produces a new_block_Square.

OK, then there is no unsoundness. I suggest we leave polymorphic destruction for next year.

> This means that destroy in the example can only be used on an object with most derived type Shape.

That means this function is pointless since Shape is abstract. Let's just remove it.

 Your two-arg `new_block` suggestion seems like a good idea! Let's do it next year.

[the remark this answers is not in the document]
btj: > OK, then there is no unsoundness.

Actually, this does not resolve the subtlety of the soundness argument.  If `delete` would check that the vtype equals the most derived type, then it would be obviously sound. I suggest we leave this issue for next year. (But if you prefer not to write text about a design that will change, feel free to do it now.)

btj:
Excellent job; many thanks!

PR #331, opened 2022-11-09
Support `typeid(struct S)` in ghost code

NielsMommen:


src/verifast1.ml
```diff
@@ -1916,20 +1917,24 @@ module VerifyProgram1(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
...
             if List.mem_assoc f fmap then static_error lf "Duplicate field name." None;
             let t = check_pure_type ("", []) [] gh t in
             let offset = if gh = Ghost then None else Some (get_unique_var_symb (sn ^ "_" ^ f ^ "_offset") intType) in
             let entry = (f, (lf, gh, t, offset, init)) in
-            iter1 (entry::fmap) fds (has_ghost_fields || gh = Ghost) bases
+            iter1 (entry::fmap) fds (has_ghost_fields || gh = Ghost) bases type_info
         in
-        let new_item = begin
+        let new_item = 
+          let type_info = get_unique_var_symb (sn ^ "_type_info") type_info_ref_type in
```
btj, line 1934: Move this line before the definition of `iter1` so you don't need the extra `iter1` parameter?

btj:
Please add a test case. Something like `//@ void *p = typeid(struct S);` (or `//@ void *p = &typeid(struct S);`?) should be possible. (If necessary, perhaps make `typeid(struct S)` be of type `void *` in C.) Otherwise, looks good! Thanks!

btj:
Don't forget to add the test case to `testsuite.mysh`!

PR #343, opened 2022-11-22
[C++] Consume bases_constructed at end of destructor body

NielsMommen:
`X_bases_constructed` was consumed too early when verifying a destructor, which didn't allow to call methods during destruction.


btj:
Did you forget to add the new test case to `testsuite.mysh`?

PR #344, opened 2022-11-25
[C++] virtual methods

NielsMommen:


tests/cxx/virtual_methods/virtual_call_in_ctor_ub.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+struct BVirtual
+{
+  int m_b;
+
+  BVirtual(AVirtual *av) : m_b(av->m_a)
+  //@ requires av != 0 &*& av->m_a |-> ?a;
+  //@ ensures av->m_a |-> a &*& BVirtual_vtype(this, &typeid(struct BVirtual)) &*& this->m_b |-> a;
+  {
+    av->foo(); //~
```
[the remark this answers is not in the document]
NielsMommen: When verifying a virtual method implementation for struct `S`, a `S_vtype` chunk is consumed (like we do for `S_bases_constructed` when `S` derives from at least one struct) and reproduced before the body is executed. Hence, `foo` requires that chunk in its precondition.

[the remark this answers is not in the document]
btj: I see. I had expected that the existence of the `S_vtype` chunk would be checked at the call site (in analogy with what happens at run time: the vtable pointer is dereferenced at the call site to get the pointer to the function to call), but I guess that indeed, at verification time, the call site does not care about the value of the vtable pointer since it is always the statically resolved spec that is used to verify it, so the vtable pointer existence check might as well happen in the callee. OK; looks good!

tests/cxx/virtual_methods/virtual_methods.cpp
```diff
@@ -0,0 +1,116 @@
...
+  /*@ ensures 
+      B_bases_constructed(this) &*&
+      A_vtype(this, t) &*& 
+      A_m_i(this, i);
+  @*/
+  {
+    callVirtualMeth();
+  }
+  
+  void callStaticlyBoundCall() const
```
btj, line 81: typo: should be `statically`

btj:
I think merging #346 caused a conflict with this PR :-/

btj:
There currently is no example where a virtual method is overridden? Best to add such an example, including one where a method overrides a virtual method from a base class that is not the first base class (i.e. where `this` is at a different address)?

NielsMommen:
> There currently is no example where a virtual method is overridden? Best to add such an example, including one where a method overrides a virtual method from a base class that is not the first base class (i.e. where `this` is at a different address)?

I didn't add such a test because the behavioral subtyping check won't succeed because the overriding method in class `B` would require `B_bases_constructed` in its precondition. I can move the check for `bases_constructed` from the callee site to the call site to allow overrides.

src/verifast1.ml
```diff
@@ -5917,18 +5941,59 @@ module VerifyProgram1(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
...
+      Java-> check_classnamelist (pn,ilist) is
+    | _ -> check_funcnamelist is 
+    in
+    check_predinst (pn, ilist) [] [] [] l pred tparams fns params body
+  
+  let predinstmap1 =
+    match dialect with 
+    | Some Cxx ->
+      (* 
+        For each polymorphic struct S with bases [B0; ...; Bn] and n > 0 create:
```
btj, line 5961: `with bases` -> `with polymorphic bases`

src/verifast1.ml
```diff
@@ -5917,18 +5941,59 @@ module VerifyProgram1(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
...
       match ds with
         PredFamilyInstanceDecl (l, p, tparams, is, xs, body)::ds ->
-        let fns = match file_type path with
-          Java-> check_classnamelist (pn,ilist) is
-        | _ -> check_funcnamelist is 
-        in
-        let (pfns, info) as entry = check_predinst (pn, ilist) [] [] [] l p tparams fns xs body in
-        let _ = if List.mem_assoc pfns pm || List.mem_assoc pfns predinstmap0 then static_error l "Duplicate predicate family instance." None in
-        iter (pn,ilist) (entry::pm) ds
+          let (pfns, info) as entry = mk_pred_inst l (pn, ilist) p tparams is xs body in
```
btj, line 5994: Bad indent

btj (changes requested):
Excellent work, Niels! Just two tiny nitpicks...

btj (approved):

PR #351, opened 2022-12-11
Forbid negative-size objects

jpdoyle:
I couldn't figure out how to prove:

```
lemma void integers__limits(void *array);
    requires [?f]integers_(array, ?sz, ?sgn, ?n, ?cs) &*& n > 0;
    ensures [f]integers_(array, sz, sgn, n, cs) &*& (void *)0 <= array
        &*& array + n <= (void *)UINTPTR_MAX
        &*& array + n*sz <= (void *)UINTPTR_MAX;
```

because there didn't seem to be any axiom guaranteeing `sz >= 1`. This seems like a reasonable spot to add that axiom.

### DIFF
--- a/bin/prelude.h
+++ b/bin/prelude.h
@@ -96,7 +96,7 @@ fixpoint bool pointer_within_limits(void *p) {
 }
 
 fixpoint bool object_pointer_within_limits(void *p, int size) {
-    return pointer_within_limits(p) && pointer_within_limits(p + size) && (uintptr_t)p != 0 && size > 0;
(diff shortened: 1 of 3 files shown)


btj:
The proposed change is inconsistent with the fact that VeriFast currently produces `object_pointer_within_limits(result, size)` after a successful `malloc` or `calloc` with size zero. The C standard seems to allow implementations to return a pointer that is not a null pointer in that case. In fact, I just tested this on MacOS and in Linux and in both cases `malloc` returns a non-null pointer.

It would definitely be fine to assume that the `size` argument of an `integer_` chunk is greater than zero, though.

PR #353, opened 2023-01-31
Delete polymorphic objects on the heap

NielsMommen:


tests/cxx/delete/non_polymorphic_base.cpp
```diff
@@ -0,0 +1,58 @@
...
+  delete a;
+}
+
+int main()
+//@ requires true;
+//@ ensures true;
+{
+  B* b = new B;
+  //@ open b->valid();
+  //@ upcast_new_block((struct A*) b);
```
btj, line 54: Shouldn't VeriFast reject this? If B had any fields, they would be deallocated by the `delete` call at run time, but VeriFast would unsoundly not consume them?

src/verifast.ml
```diff
@@ -533,6 +533,26 @@ module VerifyProgram(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
...
+        begin fun cont ->
+          verify_expr false h env None derived_w cont econt
+        end @@ fun h env derived_addr ->
+        begin fun cont ->
+          verify_expr false h env None w cont econt
+        end @@ fun h env base_addr ->
+        let derived_symb = get_pred_symb_from_map derived new_block_pred_map in
+        let base_symb = get_pred_symb_from_map base new_block_pred_map in
+        let _, _, _, _, _, _, is_virtual = List.assoc base cxx_dtor_map in
+        if not is_virtual then static_error l "The target object must have a virtual destructor." None;
```
btj, line 550: `target object` -> `target struct`?

src/verifast.ml
```diff
@@ -533,6 +533,26 @@ module VerifyProgram(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
...
+      | CastExpr (_, _, Upcast (derived_w, PtrType (StructType derived), PtrType (StructType base))) ->
+        begin fun cont ->
+          verify_expr false h env None derived_w cont econt
+        end @@ fun h env derived_addr ->
+        begin fun cont ->
+          verify_expr false h env None w cont econt
+        end @@ fun h env base_addr ->
+        let derived_symb = get_pred_symb_from_map derived new_block_pred_map in
+        let base_symb = get_pred_symb_from_map base new_block_pred_map in
+        let _, _, _, _, _, _, is_virtual = List.assoc base cxx_dtor_map in
```
btj, line 549: Can this crash if the target struct does not have a dtor?

src/verifast.ml
```diff
@@ -650,13 +670,20 @@ module VerifyProgram(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
               consume_chunk rules h [] [] [] l (uninit_array_pred_symb, true) [] real_unit real_unit_pat (Some 2) [TermPat addr; TermPat n; dummypat] @@ fun _ h _ _ _ _ _ _ ->
```
btj, line 670: BTW: I notice now that the whole CxxDelete implementation is indented too much.

src/verifast.ml
```diff
@@ -650,13 +670,20 @@ module VerifyProgram(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
...
+                    consume_chunk rules h [] [] [] l (new_block_symb, true) [] real_unit real_unit_pat (Some 1) [TermPat addr; dummypat] @@ fun _ h _ [_; type_info] _ _ _ _ -> 
+                    consume_chunk rules h [] [] [] l (get_pred_symb_from_map name cxx_vtype_map, true) [] real_unit real_unit_pat (Some 1) [TermPat addr; TermPat type_info] @@ fun _ _ _ _ _ _ _ _ ->
+                    cont h
+                  else
+                    consume_chunk rules h [] [] [] l (new_block_symb, true) [] real_unit real_unit_pat (Some 1) [TermPat addr] @@ fun _ h _ _ _ _ _ _ -> 
+                    cont h
+                end @@ fun h ->
+                consume_cxx_object l real_unit_pat addr t verify_dtor_call false h env cont
               | _ -> 
                 consume_chunk rules h [] [] [] l (get_pred_symb "new_block", true) [] real_unit real_unit_pat (Some 1) [TermPat addr; TermPat (sizeof l t)] @@ fun _ h _ _ _ _ _ _ ->
```
btj, line 688: If the type does not have a pointee chunk and is not a struct, you consume only the new_block?

btj: An example of such a type is `int32_t` in targetless mode, which is represented by an `integer_` chunk.

btj (changes requested):
Good job! Just a few minor issues.

PR #354, opened 2023-02-16
Identifier module for functions

NielsMommen:
Declared function can now have a unique identifier and a 'human readable name' at the same time. The identifier is used internally and the name is there for the user. For `C++`, mangled names are no longer displayed:
image


btj:
It seems better to define `Identifier.t` as `{id: string; display_name: string}` and expose this definition so that you can simply write `g.id` to get the identifier for `g` (no parentheses required) instead of having to write `let fn_id = Identifier.id g in ... fn_id ...`. (This also avoids having to remember the connection between g and fn_id when reviewing code.) No information is lost: an identifier g is mangled iff `g.id <> g.display_name`. This also has better performance because you avoid the extra case analysis.

btj:
But actually, wouldn't it be even better if we could construct identifiers for C++ functions that would be both unique and human-readable? E.g. `foo(int, int)` for a function overload that takes two ints? I would expect that Clang would be able to provide that. What does Clang use in its error messages?

PR #356, opened 2023-02-22
[C++] Unique readable function names

NielsMommen:
Function names include parameter types for `C++`.


src/verify_expr.ml
```diff
@@ -423,6 +423,7 @@ module VerifyExpr(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
             cont (fn, FuncInfo ([], fterm, l, k, tparams, rt, xmap, nonghost_callers_only, pre, pre_tenv, post, terminates, functype_opt, body', is_virtual)) ((fn, l0)::prototypes_implemented)
         end @@ fun func_info protos_implemented ->
         let check_override overridden_meth =
+          let () = print_endline overridden_meth in
```
btj, line 426: The surgeon left his instruments in the patient!

btj:
I have a few concerns:
- For each function, VeriFast generates a symbol that represents the function's address. Redux doesn't mind if symbol names have spaces or punctuation in them, but Z3 might. I suggest you run at least one test case with `-prover z3v4.5`. (This also does type-checking on terms, which Redux does not do.) (If it turns out Z3 fails on this, we might still decide to not support Z3 for C++ for now, but at least it's good to know. But it seems that it should be easy to fix: some straightforward scheme for replacing the non-allowed chars with some allowed char (ideally one that is not allowed in C++ identifiers) (while keeping the symbol as readable as possible, since it appears in the symbolic states) can be applied.)
- There are a few scenarios where the user has to mention a function name in an annotation. Perhaps the most important one is as an index for a predicate family (e.g. when starting a thread; see the Tutorial). It's fine if this doesn't work for overloaded functions, but perhaps we should check that works even in C++ programs if the function is not overloaded? Would it make sense to add a test case for this? (OTOH, I guess this did not work before either, so it's fine to leave this for a future improvement. Only do it now if it is a regression and/or it is easy to do.) (I guess C++ does not allow writing `void *p = &f;` if `f` is an overloaded function? Is there a way to get a pointer to an overloaded function in C++? Perhaps if you are assigning to a variable of a particular function type, it performs overload resolution? In real code, Clang will take care of this; in ghost code, it's our job, but this seems like a very low-priority issue.)

btj:
For the second issue: one solution is to extend the VeriFast lexer to support some syntax for identifiers with special characters (e.g. `` `foo(int, int)` ``).

NielsMommen:
> * For each function, VeriFast generates a symbol that represents the function's address. Redux doesn't mind if symbol names have spaces or punctuation in them, but Z3 might. I suggest you run at least one test case with `-prover z3v4.5`. (This also does type-checking on terms, which Redux does not do.) (If it turns out Z3 fails on this, we might still decide to not support Z3 for C++ for now, but at least it's good to know. But it seems that it should be easy to fix: some straightforward scheme for replacing the non-allowed chars with some allowed char (ideally one that is not allowed in C++ identifiers) (while keeping the symbol as readable as possible, since it appears in the symbolic states) can be applied in `z3v4dot5prover.ml`.)

Looks like the spaces are not problematic. I could ofcourse also remove the spaces between a `,` and a parameter.

> * There are a few scenarios where the user has to mention a function name in an annotation. Perhaps the most important one is as an index for a predicate family (e.g. when starting a thread; see the Tutorial). It's fine if this doesn't work for overloaded functions, but perhaps we should check that works even in C++ programs if the function is not overloaded? Would it make sense to add a test case for this? (OTOH, I guess this did not work before either, so it's fine to leave this for a future improvement. Only do it now if it is a regression and/or it is easy to do.) (I guess C++ does not allow writing `void *p = &f;` if `f` is an overloaded function? Is there a way to get a pointer to an overloaded function in C++? Perhaps if you are assigning to a variable of a particular function type, it performs overload resolution? In real code, Clang will take care of this; in ghost code, it's our job, but this seems like a very low-priority issue.)

Indeed, overload resolution is performed and there must exist a conversion from the function type to the *target* type. [Here](https://en.cppreference.com/w/cpp/language/overloaded_address) are some examples.

btj:
> Looks like the spaces and commas are not problematic.

Great! Then this is resolved and no further action is needed here.

> Indeed, overload resolution is performed and there must exist a conversion from the function type to the _target_ type. [Here](https://en.cppreference.com/w/cpp/language/overloaded_address) are some examples.

Thanks for the link! So, what about mentioning function names in annotations?

NielsMommen:
> Thanks for the link! So, what about mentioning function names in annotations?

I'll have a look at it now. I guess the backticks make sense to be able to parse function names with their parameters.

btj:
Yeah, actually, never mind. I'm pretty sure it didn't work before, so no need to fix it in this PR. I merged it; good job! The backticks are probably the way to go, but there is no rush to implement it now.

PR #363, opened 2023-06-01
Use new vfdeps

NielsMommen:
Updated to use new vfdeps and new parser syntax. (`depend` target is a lot faster now because we don't have to preprocess it with `camlp4`)


btj:
Great job! Many thanks for this very big effort!

PR #369, opened 2023-09-28
Simplify pointers in redux

NielsMommen:
Simplify nested `field_ptr_parent(field_ptr(target, _, offset), offset)` to `target` in redux.


btj:
Cool!!

PR #399, opened 2024-01-04
[Rust] Support for mutable references

Nima-Rahimi-Foroushaani:


btj (changes requested):
Looks good. Just a few minor changes requested. Note: most of the code is for supporting references in safe function signatures to various types other than structs and is not being tested, and probably mostly does not work yet (due to missing borrow_content predicates, if nothing else). OK to leave it in for now, but for the future, it seems better to just put `failwith "TODO"` in those cases and only write any code for it when a test case requires it.

btj (changes requested):
Looks good! Just fix the indentation issues in `purely_unsafe/deque.rs`.

btj (approved):

PR #416, opened 2024-01-18
[Rust] thread local masks

Nima-Rahimi-Foroushaani:


bin/rust/rust_belt/lifetime_logic.gh
```diff
@@ -42,7 +45,17 @@ lemma void close_lifetime_intersection_dead_token(lifetime_t k, lifetime_t k1);
...
+fixpoint void *Nshr() { return (void *) 0; }
+fixpoint void *Nrust() { return (void *) 1; }
+
+predicate thread_token(thread_id_t t, mask_t m);
+lemma void thread_token_split(thread_id_t t, mask_t m , mask_t to_split);
+    //@ requires thread_token(t, m) &*& mask_is_sub(to_split, m) == true;
+    //@ ensures thread_token(t, mask_diff(m, to_split)) &*& thread_token(t, to_split);
+
+lemma void thread_token_merge(thread_id_t t, mask_t m , mask_t m1);
+    //@ requires thread_token(t, m) &*& thread_token(t, m1) &*& mask_is_empty(mask_isect(m, m1)) == true;
```
btj, line 57: I'm pretty sure that this is_empty requirement is superfluous, so we should drop it. Does it not follow from `thread_token(t, m) &*& thread_token(t, m1)` that `mask_is_empty(mask_isect(m, m1))`?

bin/rust/rust_belt/lifetime_logic.gh
```diff
@@ -76,24 +89,24 @@ lemma void full_borrow_implies(lifetime_t k, predicate() P, predicate() P1);
     ensures full_borrow(k, P1) &*& is_implies(f, P, P1) &*& is_implies(f1, P1, P);
 
 // Non-atomic persistent borrow
-predicate nonatomic_borrow(lifetime_t k, thread_id_t t, predicate() P);
+predicate nonatomic_borrow(lifetime_t k, thread_id_t t, mask_t m, predicate() P);
 
-lemma void full_borrow_into_nonatomic_borrow(lifetime_t k, thread_id_t t, predicate() P);
+lemma void full_borrow_into_nonatomic_borrow(lifetime_t k, thread_id_t t, predicate() P, mask_t m);
```
btj, line 94: I'm pretty sure this mask must not be empty!! Same bug as before?

tests/rust/safe_abstraction/deque.rs
```diff
@@ -230,8 +230,8 @@ impl Deque {
     }
 
     pub fn pop_front<'a>(&'a mut self) -> i32
-        //@ req thread_token(?_t) &*& [?_q]lifetime_token(?a) &*& full_borrow(a, Deque_full_borrow_content(_t, self));
-        //@ ens thread_token(_t) &*& [_q]lifetime_token(a);
+        //@ req thread_token(?_t, ?m) &*& [?_q]lifetime_token(?a) &*& full_borrow(a, Deque_full_borrow_content(_t, self));
```
btj, line 233: I think it is more instructive for the reader to say TopMinus([]) here? But this makes me wonder if we should not introduce a separate predicate partial_thread_token(t, m) and define thread_token(t) = partial_thread_token(t, MaskTopMinus([])) so that people who don't know or care about nonatomic borrows are not bothered with these masks?

tests/rust/safe_abstraction/cell_u32.rs
```diff
@@ -92,5 +100,33 @@ impl CellU32 {
...
+            let tmp = *po;
+            *po = *ps;
+            *ps = tmp;
+        }
+        //@ assert thread_token(_t, ?rem);
+        //@ close_nonatomic_borrow();
+        //@ close_nonatomic_borrow();
+        //@ thread_token_merge(_t, rem, mo);
+        //@ assert thread_token(_t, ms) &*& thread_token(_t, ?rem1);
+        //@ thread_token_merge(_t, rem1, ms);
```
btj, line 130: Very nice! The fact that this proof goes through shows that mask.gh is fine computationally.

bin/rust/rust_belt/lifetime_logic.gh
```diff
@@ -42,7 +45,17 @@ lemma void close_lifetime_intersection_dead_token(lifetime_t k, lifetime_t k1);
     requires [_]lifetime_dead_token(k);
     ensures [_]lifetime_dead_token(lifetime_intersection(k, k1));
 
-predicate thread_token(thread_id_t t); // Non-atomic token with Top mask ([NaInv: t.Top] in RustBelt)
+fixpoint void *Nshr() { return (void *) 0; }
+fixpoint void *Nrust() { return (void *) 1; }
+
+predicate thread_token(thread_id_t t, mask_t m);
```
btj, line 51: Should we add a lemma that allows one to replace the mask_t value in a thread_token by an equivalent value? This way canonicity is not crucial.

btj (changes requested):
Looks good! The only required change is to fix the unsoundness in the lemma for creating a nonatomic borrow. I would also introduce the partial_thread_token predicate. Furthermore, I would also add the missing normalize calls because it's so easy and beneficial. The canonicity stuff can wait.

btj:
BTW: It might be nice to introduce a fixpoint `fixpoint mask_t MaskTop() { return MaskTopMinus({}); }` (which can be called without parentheses: `MaskTop`) and similarly for bot.

bin/rust/rust_belt/lifetime_logic.gh
Nima-Rahimi-Foroushaani: Yes indeed

tests/rust/safe_abstraction/deque.rs
Nima-Rahimi-Foroushaani: I chose the second solution. safe_abstraction/deque.rs and tree.rs are back to using `thread_token(t)`

bin/rust/rust_belt/lifetime_logic.gh
Nima-Rahimi-Foroushaani: It should follow. I simply put it there because of the disjoint union in NAInv-tok-split rule in RustBelt.

bin/mask.gh
```diff
@@ -0,0 +1,123 @@
...
+fixpoint mask_t mask_union(mask_t m, mask_t m1) {
+  switch(m) {
+    case MaskTopMinus(mns): return switch(m1) {
+      case MaskTopMinus(mns1): return MaskTopMinus(map((mask_node_isect_list)(mns1), mns));
+      case MaskBotPlus(mns1): return MaskTopMinus(mask_node_list_normalize(map((mask_node_diff_list)(mns1), mns)));
+      case MaskEmp: return m;
+    };
+    case MaskBotPlus(mns): return switch(m1) {
+      case MaskTopMinus(mns1): return MaskTopMinus(map((mask_node_diff_list)(mns), mns1));
+      case MaskBotPlus(mns1): return MaskBotPlus(mask_node_list_merge(append(mns, mns1)));
```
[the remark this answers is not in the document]
Nima-Rahimi-Foroushaani: I think the `mask.gh` preserves canonicity now. I cared to hold the strict order of nodes and not introduce empty nodes or `MaskEmp`.

bin/mask.gh
```diff
@@ -0,0 +1,117 @@
...
+  ensures mask_node_list_merge(l, l1) == switch(l) {
+    case nil: return l1;
+    case cons(h, t): return switch(l1) {
+      case nil: return l;
+      case cons(h1, t1): return fst(h) == fst(h1) ? cons(pair(fst(h), mask_union(snd(h), snd(h1))), mask_node_list_merge(t, t1)) :
+        ptr_lt(fst(h), fst(h1)) ? cons(h, mask_node_list_merge(t, l1)) : cons(h1, mask_node_list_merge(l, t1));
+    };
+  };
+
+// This computes mn / mns
```
btj, line 49: Isn't set difference denoted by a backslash? A slash denotes a quotient, which is something else.

btj (changes requested):
Very nice! Just replace the slash by a backslash and rebase and it can land!

bin/mask.gh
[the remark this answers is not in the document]
btj: Yep, looks good!

btj (approved):

PR #422, opened 2024-01-25
[Rust] Adds support for fractured borrows

Nima-Rahimi-Foroushaani:


btj:
That was nice and fast!

PR #430, opened 2024-02-07
T_own takes a T value, even if T is a struct

btj:
Also implemented struct expressions in the Rust annotation parser.
Horrible hack: if an identifier I is followed by a left brace,
this is parsed as a struct expression if and only if I starts with
an uppercase letter. (I don't know how the official Rust parser
parses this; I suspect it first resolves I to see whether it refers
to a struct!)

To have I be parsed as a variable name, enclose it in parentheses: (I).


btj:
Closing this PR; I now believe T_own should keep taking the field values.

PR #434, opened 2024-02-09
Rust parser improvements

Nima-Rahimi-Foroushaani:
These are specifically necessary to support `Mutex`.


src/rust_frontend/vf_mir_translator/rust_parser.ml
```diff
@@ -8,9 +9,10 @@ let is_expr_with_block = function
 | SwitchExpr (_, _, _, _) -> true
 | _ -> false
 
-let rec parse_type_path_rest x = function%parser
-  [ (_, Kwd "::"); (_, Ident xx); [%let x = parse_type_path_rest (x ^ "::" ^ xx)] ] -> x
-| [ ] -> x
+let rec parse_type_path_rest l x = function%parser
```
btj, line 12: Should `parse_type_path_rest` be renamed to `parse_path_rest` (or `parse_simple_path_rest`?

btj: Also, insert something into your parser_test.rs that exercises this?

btj (changes requested):
Very nice fix for the `[?q](*p)->f |-> _` parsing issue! Just a few minor changes requested.

tests/rust/parser_test.rs
```diff
@@ -1,10 +1,15 @@
...
+mod foo {
+    struct Foo {
+        v: i32,
+    }
 
-unsafe fn f(p: *mut Foo)
-//@ req [?q](*p).v |-> ?v;
-//@ ens [q](*p).v |-> v;
-{
+    //@ pred foo::Foo_own(t: thread_id_t, v: i32) = true;
```
btj, line 7: Sorry to be a nitpick, but I would prefer if you moved this definition out of the module; the natural behavior would be for this `foo::` prefix to be added automatically if the predicate is inside the `mod`. Moving it out of the `mod` makes this test case future-proof.

btj (changes requested):

tests/rust/parser_test.rs
Nima-Rahimi-Foroushaani: Good point!

btj (approved):

PR #437, opened 2024-02-13
[Rust] MutexU32

Nima-Rahimi-Foroushaani:
Interpretation of `MutexU32` and its well-formedness proof obligations pass through.


bin/rust/rust_belt/lifetime_logic.gh
```diff
@@ -78,6 +102,21 @@ lemma void close_full_borrow(predicate() P);
...
     ensures full_borrow(k, P) &*& [q]lifetime_token(k);
 
+// LftL-bor-acc-strong
+predicate close_full_borrow_token_strong(lifetime_t k1, predicate() P, real q, lifetime_t k);
+
+lemma void open_full_borrow_strong(lifetime_t k, predicate() P, real q);
+    requires full_borrow(k, P) &*& [q]lifetime_token(k);
+    ensures exists(?k1) &*& lifetime_inclusion(k, k1) == true &*& P() &*& close_full_borrow_token_strong(k1, P, q, k);
+
+typedef lemma void full_borrow_convert_strong(predicate() Q, lifetime_t k1, predicate() P)();
```
btj, line 112: Add a TODO saying that in RustBelt this viewshift is at mask Bot so certain lemmas must not be called to prove this lemma. (For example, RustBelt does not support converting `[_]lifetime_dead_token(k1)` to `[_]lifetime_dead_token(k)` inside this proof.)

btj: One way to enforce this would be to introduce a chunk `current_mask(mask_t)` which is in the pre and post of *all* Rust functions and all lemmas. Another way is to have two versions of each lemma: one that is `nonghost_callers_only`, and one that consumes and produces a `current_mask`.

bin/rust/rust_belt/lifetime_logic.gh
```diff
@@ -88,6 +127,11 @@ lemma void end_reborrow(lifetime_t k1, lifetime_t k, predicate() P);
     requires end_reborrow_token(k1, k, P) &*& [_]lifetime_dead_token(k1);
     ensures full_borrow(k, P);
 
+// LftL-bor-fake
+lemma void full_borrow_fake(lifetime_t k, predicate() P);
+    requires lifetime_dead_token(k);
```
btj, line 132:
```suggestion
    requires [_]lifetime_dead_token(k);
```

bin/rust/rust_belt/lifetime_logic.gh
```diff
@@ -145,4 +195,16 @@ lemma void close_frac_borrow(real q_p, predicate(;) P);
     requires [q_p]P() &*& close_frac_borrow_token(q_p, P, ?q_lft, ?k);
     ensures [q_lft]lifetime_token(k);
 
+// LftL-fract-acc-atomic
+predicate close_frac_borrow_token_atomic(bool b, real q, predicate(;) P, lifetime_t k);
+
+lemma void open_frac_borrow_atomic(lifetime_t k, predicate(;) P);
```
btj, line 201: This obviously needs to consume something like a `current_mask` chunk.

src/verifast1.ml
```diff
@@ -2713,6 +2713,7 @@ module VerifyProgram1(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
   let rec type_satisfies_contains_any_constraint allowContainsAnyPositive tp =
     match unfold_inferred_type tp with
       Bool | AbstractType _ | Int (_, _) | Float | Double | LongDouble | RealType | FuncType _ | PtrType _ | ObjType _ | ArrayType _ | BoxIdType | HandleIdType -> true
+(*    | StructType (_, ts) -> List.for_all (fun t -> type_satisfies_contains_any_constraint allowContainsAnyPositive t) ts *)
```
btj, line 2716: A struct type `struct S<T1,...,Tk> { f1: t1, ..., fN: tN }` should be treated like an `inductive S<T1,...,Tk> = S(t1, ..., tN);`

tests/rust/safe_abstraction/mutex.rs
```diff
@@ -0,0 +1,241 @@
...
+        impl Mutex {
+            pub unsafe fn new() -> Mutex
+            //@ req true;
+            //@ ens SysMutex(result, MutFree);
+            {
+                abort();
+            }
+
+            pub unsafe fn lock<'a>(&'a self)
+            //@ req thread_token(?t) &*& [?q](*self) |-> ?m &*& SysMutex(m, _);
```
btj, line 37: This function requires exclusive ownership of the SysMutex chunk??

btj (commented):
Let's discuss...

btj (commented):
I finished reviewing everything except for mutex_u32.rs. Looks good; just a few minor comments.

tests/rust/safe_abstraction/mutex_u32.rs
```diff
@@ -0,0 +1,444 @@
+#![feature(negative_impls)]
+#![allow(dead_code)]
+
+/*
+About the definitions:
+    Since in the proof of `MutexU32::lock` we need to close the `frac_borrow`,
+    in the `sys::locks::Mutex` interface; i.e. `SysMutex` fractions required by `sys::locks::Mutex::lock` to lock the mutex should be turened back from the method,
+    we cannot keep track of `SysMutex` fractions used to get the lock. So we adapth an interface for `sys::locks::Mutex` such that it does not care
+    if the mutex is locked or not.
+*/
```
btj, line 10: Please check for correct English spelling and grammar throughout this file. I have trouble parsing this sentence.

tests/rust/safe_abstraction/mutex_u32.rs
```diff
@@ -0,0 +1,444 @@
...
+
+/*@
+// dummy
+pred sys::locks::Mutex_own(t: thread_id_t, m: *u32);
+pred sys::locks::Mutex_share(k: lifetime_t, t: thread_id_t, l: *sys::locks::Mutex;) = true;
+lem sys::locks::Mutex_share_mono(k: lifetime_t, k1: lifetime_t, t: thread_id_t, l: *sys::locks::Mutex)
+req lifetime_inclusion(k1, k) == true &*& [_]sys::locks::Mutex_share(k, t, l);
+ens [_]sys::locks::Mutex_share(k1, t, l);
```
btj, line 18: Please indent.

tests/rust/safe_abstraction/mutex_u32.rs
```diff
@@ -0,0 +1,444 @@
...
+{}
+lem sys::locks::Mutex_share_full(k: lifetime_t, t: thread_id_t, l: *sys::locks::Mutex)
+req full_borrow(k, sys::locks::Mutex_full_borrow_content(t, l)) &*& [?q]lifetime_token(k);
+ens [_]sys::locks::Mutex_share(k, t, l) &*& [q]lifetime_token(k);
```
btj, line 22: Please indent.

tests/rust/safe_abstraction/mutex_u32.rs
```diff
@@ -0,0 +1,444 @@
...
+pred SysMutex(m: sys::locks::Mutex; P: pred());
+pred SysMutex_share(l: *sys::locks::Mutex; P: pred());
+lem SysMutex_share_full(l: *sys::locks::Mutex);
+    req *l |-> ?m &*& SysMutex(m, ?P);
+    ens SysMutex_share(l, P);
+lem SysMutex_end_share(l: *sys::locks::Mutex);
+   req SysMutex_share(l, ?P);
+   ens *l |-> ?m &*& SysMutex(m, P);
+
+pred SysMutex::new_ghost_arg(P: pred()) = true;
```
btj, line 37: The modern way is to just use `exists`. (I created `threading.h` before I had thought of `exists` as a general pattern.)

tests/rust/safe_abstraction/mutex_u32.rs
```diff
@@ -0,0 +1,444 @@
...
+   So the proof sketch for this lemma would be:
+   No other thread has access to this mutex.
+   If
+       - The mutex is free and the resource `P` is protected by mutex; It is in the mutex so to speak. We substitute `P` with `Q`
+       and from this point there will be no `[q]SysMutex(m, P)` anywhere. It means we forget about `P`. The state is just like after
+       a call to `sys::locks::Mutex::new` to protect `Q()`.
+
+       - The mutex is locked it means there should be a `SysMutex_locked(m, P(), ?t, ?q)` and `P()` in resources of some thread(s).
+       Operations in this state:
+           - `lock` gets verified which is fine because this call will not terminate unless someone frees the lock. after that we get
```
btj, line 53: By "frees" you mean "unlocks"?

tests/rust/safe_abstraction/mutex_u32.rs
```diff
@@ -0,0 +1,444 @@
...
+                abort();
+            }
+
+            // Todo: impl Drop for Mutex
+            unsafe fn drop<'a>(&'a mut self)
+            //@ req (*self) |-> ?m &*& SysMutex(m, _);
+            //@ ens (*self) |-> m;
+            {
+                abort();
+            }
```
btj, line 97: This seems unsound. It seems this allows me to unlock a lock that has been dropped? I think `unlock` needs to require a fraction of SysMutex_share.

btj (changes requested):

tests/rust/safe_abstraction/mutex_u32.rs
```diff
@@ -0,0 +1,444 @@
...
+       The target lifetimes always outlive reference lifetime out of compiler guarantees of wellformedness of types.
+    */
+    @*/
+    //@ ens thread_token(t) &*& [qa]lifetime_token(a) &*& full_borrow(a, u32_full_borrow_content(t, result));
+    {
+        //@ let kstrong = open_full_borrow_strong(a, MutexGuardU32_full_borrow_content0(km, t, self), qa/2);
+        //@ open MutexGuardU32_full_borrow_content0(km, t, self)();
+        //@ open MutexGuardU32_own(km)(t, ?lock);
+
+        // We need `(*lock).data |-> _` to get ptr provenance info
```
btj, line 303: Could you instead just add the `ptr_within_range(...)` fact in the MutexGuardU32_own predicate?

tests/rust/safe_abstraction/mutex_u32.rs
```diff
@@ -0,0 +1,444 @@
...
+    One way to handle this is to generate a different contract for `drop` implementations which does not return the lifetime token corresponding to lifetime
+    of the mutable reference. It will represent the fact that we know for this special case, i.e. `drop` this external lifetime will immediately end
+    after this function call and the original value will not be used afterward too.
+    */
+    // Todo: It should be an `impl` of `Drop` and a safe function
+    unsafe fn drop<'a>(&'a mut self)
+    /*@ req thread_token(?t) &*& [?qa]lifetime_token(?a) &*& exists(?km)
+        &*& full_borrow(a, MutexGuardU32_full_borrow_content0(km, t, self)) &*& lifetime_inclusion(a, km) == true;
+    @*/
+    //@ ens thread_token(t) /* &*& [qa]lifetime_token(a) */; // read the comment above
```
btj, line 354: This spec for `drop` seems bad; the postcondition should assert the points-to (i.e. the physical storage) for `self`. Actually, I feel we should be consistent here with how `drop` is currently implemented in the translator and have the full_borrow_content for `self` in the precondition and the full_borrow_content for the fields of `self` in the postcondition.

btj: Proposal: The spec for dropping a struct that has lifetime params should require and ensure a lifetime token fraction for each lifetime parameter. (But no token for the lifetime of the ref-to-self because that's not a real borrow.)

tests/rust/safe_abstraction/mutex_u32.rs
```diff
@@ -0,0 +1,444 @@
...
+        //@ assert [_]exists_np(?kmlong);
+        //@ lifetime_token_trade(a, qa/2, km);
+        //@ assert [?qkm]lifetime_token(km);
+        //@ open_frac_borrow(km, MutexU32_frac_borrow_content(kmlong, t, lock), qkm);
+        //@ open MutexU32_frac_borrow_content(kmlong, t, lock)();
+        unsafe {
+            (*self.lock).inner.unlock();
+        }
+        //@ assert [?qp]SysMutex_share(&(*lock).inner, _);
+        //@ close [qp]MutexU32_frac_borrow_content(kmlong, t, lock)();
```
btj, line 368: With the spec you wrote, there's nothing forcing you to close the frac borrow here. Which shows that the spec you wrote needs work.

tests/rust/safe_abstraction/mutex_u32.rs
```diff
@@ -0,0 +1,444 @@
...
+        //@ close_frac_borrow(qp, MutexU32_frac_borrow_content(kmlong, t, lock));
+        //@ lifetime_token_trade_back(qkm, km);
+        //@ leak (*self).lock |-> _;
+        //@ leak close_full_borrow_token(_, _, _);
+        //@ leak [_]lifetime_token(a);
+    }
+}
+// /*
```
btj, line 376: Remove these comments.

btj (changes requested):
I read through the whole thing. The only nontrivial TODOs before we can merge this PR are to fix the spec of SysMutex::unlock and the spec of MutexGuardU32::drop, both of which should be very little work. So it looks good!

tests/rust/safe_abstraction/mutex_u32.rs
Nima-Rahimi-Foroushaani: To have the spec for `sys::locks::Mutex::new` like `req exists(?P) &*& P();` causes the following error:
Variable P is not of predicate type.
It seems the type inference sees `P` as a fixpoint that returns a unit.

btj: To work around this limitation, write `req exists::<pred()>(?P) &*& P();`.

tests/rust/safe_abstraction/mutex_u32.rs
```diff
@@ -0,0 +1,444 @@
...
+            let data = UnsafeCell::new(v);
+            let r = MutexU32 { inner, data };
+            // Todo: Dereferencing a pointer of type struct sys::locks::Mutex is not yet supported.
+            //@ close MutexU32_own(_t, inner, data);
+            r
+        }
+    */
+    /*
+    https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.lock
+    "The exact behavior on locking a mutex in the thread which already holds the lock is left unspecified.
```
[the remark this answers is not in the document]
Nima-Rahimi-Foroushaani: This raised a question for me. This is a direct quote from documentation. Should we correct direct quotes?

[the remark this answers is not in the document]
btj: No, we should not, and in fact, I now see how it is really correct English after all. So no action required here.

tests/rust/safe_abstraction/mutex_u32.rs
```diff
@@ -186,30 +196,120 @@ lem MutexU32_share_full(k: lifetime_t, t: thread_id_t, l: *MutexU32)
...
+
+lem MutexU32_send(t: thread_id_t, t1: thread_id_t)
+    req MutexU32_own(t, ?inner, ?data);
+    ens MutexU32_own(t1, inner, data);
+{
+    open MutexU32_own(t, inner, data);
+    close MutexU32_own(t1, inner, data);
+}
+
+lem MutexU32_sync(t: thread_id_t, t1: thread_id_t)
```
btj, line 208: Whoops! Big proof! Not an ideal advertisement for our approach, I guess. An alternative might be to change the definition of MutexU32_share so that it doesn't use `t` at all and instead uses some dummy default thread id as the argument for u32_fbc? (You could use `default_value<thread_id_t>()`.) Then this proof would be trivial?

tests/rust/safe_abstraction/mutex_u32.rs
```diff
@@ -0,0 +1,444 @@
...
+        }{
+            full_borrow_implies(k, MutexU32_full_borrow_content0(t, l), sep(MutexU32_fbc_inner(l), u32_full_borrow_content(t, &(*l).data)));
+        }
+        full_borrow_split_m(k, MutexU32_fbc_inner(l), u32_full_borrow_content(t, &(*l).data));
+        let kstrong = open_full_borrow_strong_m(k, MutexU32_fbc_inner(l), q);
+        produce_lem_ptr_chunk full_borrow_convert_strong(True, MutexU32_frac_borrow_content(k, t, l), kstrong, MutexU32_fbc_inner(l))() {
+            open MutexU32_frac_borrow_content(k, t, l)();
+            SysMutex_end_share(&(*l).inner);
+            assert (*l).inner |-> ?inner;
+            SysMutex_renew(inner, True);
```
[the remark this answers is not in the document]
Nima-Rahimi-Foroushaani: Very good point. I do not think these lemmas are provable at the bottom mask. I think the least Iris feature necessary for this will be invariant, actually a cancellative one.
`full_borrow_conovert_strong` should be proven at the bottom mask. Does that mean it is broken?
Another concern that comes to my mind is the `Ctx()` predicate I have added to `full_borrow_convert_strong`'s contract. One could use it to sneak a mask token in. It does not give the `Ctx()` back in `close_full_borrow_strong`'s post-condition. So it should be fine but thought it worth mentioning to you.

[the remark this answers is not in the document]
btj: To be investigated at some point in the future (probably after we implement SysMutex)...

tests/rust/safe_abstraction/mutex_u32.rs
Nima-Rahimi-Foroushaani: Adding `pointer_within_limits`  to `MutexGuardU32_own` means `MutexGuardU32_new` should provide it which in turn it means either this method should open the full borrow or should receive it by contract. the latter means `MutexU32::lock`should open the full borrow to get that fact. So I think it would not save us proof lines for now. However, If we put the fact in `MutexGuardU32_own` it will help when we implement the `deref` method later.

I have type matching problem writing `field_pte(lock, ...)` because it needs a value of `pointer` type in the position of `lock` and `lock` is just `*MutexU32`.

tests/rust/safe_abstraction/mutex_u32.rs
Nima-Rahimi-Foroushaani: It is possible to use a `pred_ctor u32_fbc(l: *u32)()` i.e. a `thread_id` independent version of `u32_full_borrow_content` and then make `MutexU32_frac_borrow_contetn` independent of `t`. but thinking about the generic version of `Mutex` there we should say that `T` is `Send`, do `[[T]].own(t, vs) -* [[T]].own(t1, vs)` and from that we can get the logical equivalence of full borrows of `T_full_borrow_content` with `t` and `t1` args. So the same proof pattern happens there and there would be the same big proof for that. Since `produce_lem_ptr_chunk` takes the chunks produced back, we need to duplicate equivalence proofs every time. If there is an alternative way to do that in VeriFast without duplication the proof would not be very long actually.

tests/rust/safe_abstraction/mutex_u32.rs
btj: One can cast a value `v` of type `T *` to type `pointer` using a cast: `(pointer)v`. But we can look at this later.

tests/rust/safe_abstraction/mutex_u32.rs
btj: I wonder if, as part of splitting the full borrow, we can rewrite `T_fbc(t, l)` to `T_fbc(default_value, l)`. But let's look at this later.

btj (approved):

PR #443, opened 2024-02-18
Add support for Character Literals in C++

CyberChrisDE:
Verifast doesn't yet support character literals in C++. This pull request should suggest a way to change this.
As the current implementation doesn't support passing an integer value to `IntLit` (through Cap'n Proto), this implementation translates the integer value to a hex string and lets it be translated back to an integer in a later parsing step in Verifast. I don't think it's hard to change this behavior but I don't see a competence for me to do such a conceptual change.


btj:
Hi Christian, Many thanks for contributing, and it's impressive that you are so familiar with VeriFast and its codebase! Could you perhaps also contribute a test case that exercises your code? Either extend an existing test case called from `/testsuite.mysh` or add a new one to `/testsuite.mysh`.

BTW: We would be interested to learn about the context of your interest in VeriFast. What are you trying to do with it?

btj:
Many thanks, Christian! And apologies for the latency.

PR #454, opened 2024-03-04
[Rust] Track close atomic masks instead of the opened ones

Nima-Rahimi-Foroushaani:


btj:
Actually, I now realize that my original `close_atomic_space` lemma was unsound. The new version is still unsound. One can create any number of invariants with the same mask. This is fine in itself, but in this case one can open one invariant, and then close some other invariant with the same mask. One solution is for `open_atomic_space` to create a `close_atomic_space_token(spaceMask, inv)` that tracks the mask and invariant of the specific opened space. It's up to you whether you prefer to fix that in this PR or else we fix it separately.

Nima-Rahimi-Foroushaani:
@btj let's fix it in this PR

PR #470, opened 2024-03-21
Iris masks for RustBelt rules

Nima-Rahimi-Foroushaani:


bin/mask.gh
```diff
@@ -13,6 +13,13 @@ fixpoint mask_t MaskTop() { return MaskTopMinus({}); }
 fixpoint mask_t MaskBot() { return MaskBotPlus({}); }
 fixpoint mask_t MaskSingle(void *l) { return MaskBotPlus({pair(l, MaskTopMinus({}))}); }
 
+fixpoint mask_t MaskLin(list<void *> ns) {
```
btj, line 16: Not sure about the name. How about `MaskNs`?

bin/rust/rust_belt/lifetime_logic.gh
```diff
@@ -126,23 +156,83 @@ Therefore the precondition does not need to be checked in the `requires` clause
...
 
 lemma void close_frac_borrow(real q_p, predicate(;) P);
+    nonghost_callers_only
     requires [q_p]P() &*& close_frac_borrow_token(q_p, P, ?q_lft, ?k);
     ensures [q_lft]lifetime_token(k);
 
+// Mask preserving view shifts - mask aware versions. See the documentation about Iris view shifts in general.h
+// LftL-begin
+lemma lifetime_t begin_lifetime_m();
+    requires atomic_mask(Nlft);
```
btj, line 186: Should we make it easier on users and require `atomic_mask(?mask) &*& mask_le(Nlft, mask)`?

bin/rust/rust_belt/lifetime_logic.gh
```diff
@@ -126,23 +156,83 @@ Therefore the precondition does not need to be checked in the `requires` clause
...
+    ensures atomic_mask(Nlft) &*& full_borrow(k, P);
+
+// LftL-bor-na
+lemma void full_borrow_into_nonatomic_borrow_m(lifetime_t k, thread_id_t t, mask_t m, predicate() P);
+    requires atomic_mask(Nlft) &*& full_borrow(k, P) &*& !mask_is_empty(m);
+    ensures atomic_mask(Nlft) &*& [_]nonatomic_borrow(k, t, m, P);
+
+// LftL-na-acc
+lemma void open_nonatomic_borrow_m(lifetime_t k, thread_id_t t, mask_t m, real q);
+    requires atomic_mask(mask_union(Nlft, m)) &*& [_]nonatomic_borrow(k, t, m, ?P) &*& [q]lifetime_token(k) &*& partial_thread_token(t, m);
```
btj, line 217: Is it necessary to require `m` in the atomic mask here?

btj (commented):
Looks good! Just a few comments. If you resolve them and mark the PR as ready, I will merge!

bin/rust/rust_belt/lifetime_logic.gh
Nima-Rahimi-Foroushaani: LftL-na-acc demands it
[https://gitlab.mpi-sws.org/iris/lambda-rust/-/blob/master/lifetime/na_borrow.v?ref_type=heads#L39](url)

bin/rust/rust_belt/lifetime_logic.gh
Nima-Rahimi-Foroushaani: That is nice for normal rules. For mask-preserving sym accessors, it is worth noting the closing part of the rule will require the mask the opening part has chosen. It might be fine to violate that. However to be safe and not track the mask the open part used in the `close_..._token`s I am enforcing the open and close with minimum mask required

bin/rust/rust_belt/lifetime_logic.gh
btj: OK, then yes, let's require it here, too. I guess it doesn't matter too much anyway, since nonatomic borrows are probably typically opened, well, nonatomically, so with top mask available.

bin/rust/rust_belt/lifetime_logic.gh
btj: Right; makes sense. OK, let's leave it like this for now; we can always tweak this later.

Nima-Rahimi-Foroushaani: Based on the conversation we had I made it now more user-friendly.

PR #491, opened 2024-04-17
[C++] Support Pointer to Integer casts

CyberChrisDE:
For clang, uintptr_t is just a typedef that may be defined by the programmer itself. To support calculations with pointer values anyway I added support for casts from pointers to integer types. The Verifast core supported this for integers with PtrRank only so far. So I had to add support for those casts more generally.

Please have a closer look at my added line in verifast1.ml to be sure that my changes are sound.
The removed line in verifast1.ml had an exact duplicate some lines below.

### DIFF
--- a/src/cxx_frontend/stubs/stubs_ast.capnp
+++ b/src/cxx_frontend/stubs/stubs_ast.capnp
@@ -446,6 +446,7 @@ struct Expr {
     operatorCall @19 :Call;
     cleanups @20 :ExprNode;
     bindTemporary @21 :ExprNode;
+    integralCast @22 :StructToStruct;
   }
 }
 
--- a/src/cxx_frontend/stubs/stubs_ast.capnp
+++ b/src/cxx_frontend/stubs/stubs_ast.capnp
@@ -418,7 +418,7 @@ struct Expr {
     type @2 :Type;
   }
 
-  struct StructToStruct {
+  struct Cast {
     expr @0 :ExprNode;
     type @1 :Type;
   }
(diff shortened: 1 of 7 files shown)


src/cxx_frontend/stubs/stubs_ast.capnp
```diff
@@ -446,6 +446,7 @@ struct Expr {
     operatorCall @19 :Call;
     cleanups @20 :ExprNode;
     bindTemporary @21 :ExprNode;
+    integralCast @22 :StructToStruct;
```
btj, line 449: Rename `StructToStruct`? Perhaps `Cast`?
```suggestion
    integralCast @22 :Cast;
```

btj (changes requested):
Looks great! Just one suggestion.

btj:
Interestingly, the Windows build fails with the following error message:
```
> D:/a/verifast/verifast/tests/cxx/casts.cpp(5,26): cast from pointer to smaller type 'unsigned long' loses information
```
Apparently, clang uses the host architecture as the target architecture, rather than whatever is specified on the VeriFast command line. For purposes of this PR, perhaps work around it by using `unsigned long long` instead of `unsigned long` in the test case. We might want to investigate this host dependency in a separate issue/PR (if someone cares enough).

CyberChrisDE:
Type 'StuctToStruct' is used at some points already. As it should be the same for all casts, I changed its name to 'Cast'.
As you correctly stated, the different size of 'unsigned long' on Windows and Linux was the problem. I simply forgot to consider compilation/verification on Windows. It's fixed now.
For me there is no need for opening an issue. For now I don't really have a need for Windows support at all. Maybe at some point in the future. If someone wishes to have cross platform support, I'd have a look if I'm able to do the necessary changes.

NielsMommen (approved):
Looks good to me!

NielsMommen (approved):

btj (approved):

PR #518, opened 2024-05-25
[C++] Switch statement

NielsMommen:


testsuite.mysh
```diff
@@ -695,6 +695,7 @@ cd tests
     verifast -c -target Linux64 literals.cpp
     verifast -c loops.cpp
     verifast -c casts.cpp
+    verifast -c -allow_dead_code switch.cpp
```
btj, line 698: Instead of using `-allow_dead_code`, it's safer to mark the specific lines that contain dead code as `//~dead_code`. That way, if VeriFast incorrectly thinks other lines are dead code as well, it will be detected.

btj (changes requested):

PR #521, opened 2024-05-31
[C++] Deleted and Defaulted functions

CyberChrisDE:
Till now deleted and defaulted functions caused segmentation violations during serialization because they are definitions but don't have a body. This pull request should make the serializer ignore deleted functions and create error messages for defaulted functions.

Deleted functions must not be used by the programmer and clang should already check that this is ensured. This should imply that we don't need to pass them to verifast and thus don't need to serialize them at all.

Defaulted functions are more complicated. I added an error message for them to replace the segmentation violations by hopefully expressive output.

@NielsMommen I think the most important part to look at is the change in NodeSerializer.h. Maybe you have an idea how to do better.

### DIFF
--- a/src/cxx_frontend/ast_exporter/NodeListSerializer.h
+++ b/src/cxx_frontend/ast_exporter/NodeListSerializer.h
@@ -33,12 +33,6 @@ public:
         m_orphanage.newOrphan<stubs::Loc>(), m_orphanage.newOrphan<Stub>());
     m_serializer.serialize(item, nodeOrphan.locOrphan.get(),
                            nodeOrphan.descOrphan.get());
-
-    //if(std::is_convertible<T, clang::Decl const *>::value
-    if constexpr(std::is_same<stubs::Decl, Stub>::value) if(nodeOrphan.descOrphan.get().isUnionNotInitialized())
-    {
-      m_orphans.pop_back();
-    }
   }
 
   template <typename T> void serialize(llvm::ArrayRef<T> container) {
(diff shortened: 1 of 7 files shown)


src/cxx_frontend/ast_exporter/NodeListSerializer.h
```diff
@@ -33,6 +33,12 @@ template <typename Stub, typename NodeSerializerImpl> class NodeListSerializer {
         m_orphanage.newOrphan<stubs::Loc>(), m_orphanage.newOrphan<Stub>());
     m_serializer.serialize(item, nodeOrphan.locOrphan.get(),
                            nodeOrphan.descOrphan.get());
+
+    //if(std::is_convertible<T, clang::Decl const *>::value
+    if constexpr(std::is_same<stubs::Decl, Stub>::value) if(nodeOrphan.descOrphan.get().isUnionNotInitialized())
+    {
+      m_orphans.pop_back();
+    }
```
NielsMommen, line 41: At this point the orphan was already constructed, which means that corresponding memory was already allocated for the orphan in the `capnp` message. The receiver of the message won't be able to decode the orphan (which is fine), but he allocated memory in the message would be wasted.

You could add a `Deleted` declaration to the declaration union in the `capnp` schema and set is as active member when a declaration is a deleted method. In `Verifast`, we can simply ignore these `Deleted` declarations by returning an empty list in `decl_translator`.

For defaulted methods it should be fine to have `UnionNotInitialized` as active union member because you already emit an error. `ast_translator` first checks if any errors were serialized before decoding the translation unit. If errors are found, decoding of the translation unit is skipped. Therefore, the `UnionNotInitialized` won't be translated.

NielsMommen (changes requested):

NielsMommen (approved):

NielsMommen (approved):

src/cxx_frontend/ast_exporter/NodeListSerializer.h
CyberChrisDE: You meant like this?

NielsMommen: Looks good! Thanks!

PR #562, opened 2024-08-29
Tutorial examples

Nima-Rahimi-Foroushaani:


examples/rust/tutorial/alloc_u8.rs
```diff
@@ -0,0 +1,15 @@
...
+//@ req true;
+//@ ens true;
+{
+    let l = Layout::new::<u8>();
+    let p = alloc(l);
+    if p.is_null() {
+        handle_alloc_error(l);
+    }
+    *p = 42;
+    assert!(*p == 42);
```
btj, line 13: Does VeriFast support the code that this macro expands to? See how I define an assert function in the examples. Or maybe we should add support for this macro.

examples/rust/tutorial/unsafe_box_u8.rs
```diff
@@ -0,0 +1,41 @@
+use std::alloc::{alloc, dealloc, handle_alloc_error, Layout};
+
+struct BoxU8 {
+    ptr: *mut u8,
+}
+
+impl BoxU8 {
+    pub unsafe fn new(v: u8) -> BoxU8
+    //@ req true;
+    //@ ens *result.ptr |-> v &*& std::alloc::alloc_block(result.ptr, std::alloc::Layout::new_::<u8>()) &*& object_pointer_within_limits(result.ptr, 1) == true;
```
btj, line 10: Is this `within_limits` business necessary here?

examples/rust/tutorial/unsafe_box_u8.rs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        Self { ptr: raw }
+    }
+
+    pub unsafe fn into_raw(this: BoxU8) -> *mut u8
+    //@ req *this.ptr |-> ?v &*& std::alloc::alloc_block(this.ptr, std::alloc::Layout::new_::<u8>()) &*& object_pointer_within_limits(this.ptr, 1) == true;
+    //@ ens *result |-> v &*& std::alloc::alloc_block(result, std::alloc::Layout::new_::<u8>()) &*& object_pointer_within_limits(result, 1) == true;
+    {
+        this.ptr
+    }
```
btj, line 40: We should probably introduce a `get` method and a `set` method instead, to encapsulate the read and write of the u8, even though these don't exist on real `Box`. We can write "For simplicity, we consider `get` and `set` methods for now; we will consider real accessor methods in Sec. X."

examples/rust/tutorial/unsafe_box_u8.rs
```diff
@@ -0,0 +1,41 @@
...
+}
```
btj, line 41: This example should also have a `main` function that reimplements the original one from the first examples using the BoxU8.

examples/rust/tutorial/alloc_u8.rs
Nima-Rahimi-Foroushaani: Yes! Fortunately, it does.

Issue #610, opened 2024-11-25
[Rust] Check struct interpretation variance


btj:
Rust sometimes treats structs as [covariant or contravariant](https://doc.rust-lang.org/reference/subtyping.html) in their generic parameters. VeriFast currently does not yet check that the `<S>.own` and `<S>.share` predicates provided by the user for a struct S respect the struct's variance.

Thanks to Ralf Jung for noticing this.

btj:
Consider, for example, the following struct definition:

```rust
struct MyBox<T> {
    ptr: *const T; // Covariant
}
```

We might want to generate a proof obligation such as the following:
```rust
/*@

lem MyBox_own_variance<T0, T1>()
    req <MyBox<T0>>.own(?t, ?v) &*& is_subtype_of(typeid(T0), typeid(T1)) == true;
    ens <MyBox<T1>>.own(t, upcast::<MyBox<T0>, MyBox<T1>>(v));
{
    // ...
}

lem MyBox_share_variance<T0, T1>()
    req [_]<MyBox<T0>>.share(?k, ?t, ?l) &*& is_subtype_of(typeid(T0), typeid(T1)) == true;
    ens [_]<MyBox<T1>>.share(k, t, l);
{
    // ...
}

@*/
```

In fact, the latter proof obligation can be merged with MyBox_share_mono.

PR #613, opened 2024-11-27
[Rust] Check variance

btj:
For each struct S for which the user defines a custom `<S>.own` predicate, and that has non-invariant generic parameters, a proof obligation S_own_mono is now generated, requiring the user to show that the type interpretation satisfies the inferred variance.

Also:
- Generates an `S_upcast` lemma defining upcast::<S<T0...>, S<T1...>>(v)
- Improves VeriFast type inference for pure functions with a generic return type (such as `upcast`)

TODO:
- Verify Rc_own_mono and Arc_own_mono
- Update the S_share_mono proof obligation to also check monotonicity of <S>.share with respect to S's type parameters.


btj:
This partially addresses #610 .

PR #640, opened 2024-12-08
Fix tests/const_vars.c

docprofsky:
This fixes the spelling of `ptr_set` so its `should_fail` passes for using a `const`-qualified location instead of a missing function.

The other changes are to let GCC and Clang with `-Wall -Wextra -pedantic` compile tests/const_vars.c without warnings about pointer type compatibility.


btj:
Oops! Good catch!

PR #677, opened 2025-01-15
Interlude

Nima-Rahimi-Foroushaani:
It allows adding `rsspec` files the same as the prelude, but with `std/lib.rsspec` in their context.


btj:
I don’t feel the use case you’re using this for here motivates this sufficiently. It’s a nice shorthand, I guess, but wouldn’t it be better to simply add it inside std/lib.rsspec? Yes, a `use` statement would be required, but that’s standard Rust practice. It’s fine to not show the `use` statements in code listings in papers.

Nima-Rahimi-Foroushaani:
Ok. Should I remove the PR or keep it in case of more use cases in future?

btj:
Let's close it for now; it can always be found among the closed PRs?

PR #699, opened 2025-01-28
Preserve struct data when converting to byte array

NielsMommen:


src/verifast.ml
```diff
@@ -679,7 +716,15 @@ module VerifyProgram(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
       end $. fun h elems ->
       let init =
         match name with
-          "close_struct" -> Unspecified
+        | "close_struct" ->
+          begin match List.assoc sn structmap with
+          | _, _, Some (_, fields_map, _), _, _ -> 
+            let of_bytes_symb = get_pure_func_symb (match dialect with Some Rust -> "of_u8s_" | _ -> "of_chars_") in
+            let (_, _, _, _, csym_opt) = List.assoc sn struct_accessor_map in
```
btj, line 723: This line is superfluous?

src/verify_expr.ml
```diff
@@ -1595,7 +1595,7 @@ module VerifyExpr(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
           Default -> cont h env (Some (match tp with PtrType _ -> null_pointer_term () | Bool -> false_term | _ -> ctxt#mk_intlit 0))
         | Expr e -> eval_h h env e $. fun h env value -> cont h env (Some value)
         | Unspecified -> cont h env None
-        | Term t -> cont h env (Some t)
+        | Term t | MaybeUninitTerm t -> cont h env (Some t)
```
btj, line 1598: This doesn't make sense to me. Please explain? It seems better to simply not handle the `MaybeUninitTerm` case here?

src/verify_expr.ml
```diff
@@ -1841,30 +1858,41 @@ module VerifyExpr(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
...
-          cont chunks h (Some (if consumeUninitChunk then real_unit (* dummy term; should never be used *) else ctxt#mk_app csym (List.rev vs)))
+          let (_, csym, _, _, csym_opt) = List.assoc sn struct_accessor_map in
+          let csym_symb = if consumeUninitChunk then csym_opt else csym in
+          let struct_val = ctxt#mk_app csym_symb (List.rev vs) in
+          cont chunks h (Some struct_val)
         | (f, (lf, gh, t0, offset, finit))::fields ->
           let t = instantiate_type tpenv t0 in
+          let prover_convert_term_as_option_types term =
+            prover_convert_term term (option_type t) (option_type t0) 
+          in
```
btj, line 1869: This is pointless: all option types are of ProvertypeInductive, so no conversion required.

src/verify_expr.ml
```diff
@@ -1774,11 +1788,14 @@ module VerifyExpr(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
...
-            end $. fun h env value ->
-            assume_field h env sn tparams f t0 targs gh addr RegularPointsTo value coef $. fun h ->
+              | Expr e -> eval_h h env e (fun h env v -> cont h env t0 (Some v))
+              | Term t -> cont h env t0 (Some t)
+              | MaybeUninitTerm t -> 
+                let t = prover_convert_term t t0 (instantiate_type tpenv t0) in
+                cont h env (option_type t0) (Some t)
+              | Unspecified -> cont h env t0 (match gh with Ghost -> Some (get_unique_var_symb_ "value" t true) | Real -> None)
+            end $. fun h env t0 value ->
+            assume_field h env sn tparams f t0 targs gh addr pointsto_kind value coef $. fun h ->
```
btj, line 1798: It seems wrong to pass the option type as the field type? assume_field is aware that if the pointsto_kind is MaybeUninit, then the value is an option of the field type.

btj: Except that there does seem to be a bug in assume_field: it will assume_bounds of tv at frange even if pointsto_kind is MaybeUninit. But I think it's better to fix that bug than to pass the option type as the field type. If pointsto_kind is MaybeUninit and tv is Some tv, then assume_field should assume_bounds tv (option_type frange). (Currently assume_bounds does nothing for option types, but we might want to have it assume the bounds of the value inside in the future.)

btj (changes requested):
Very nice!! Many thanks! I just have a few comments and questions.

src/verify_expr.ml
NielsMommen: In Rust, `produce_points_to()` is used when closing a struct that is not a member of `structmap`. Hence, `MaybeUninitTern (u8s__of(...))` is passed to `produce_c_object` when closing it.

btj: But the code you're proposing will in that case produce a non-underscore (initialized) points-to chunk whose value is a ctor_opt value? That seems wrong? (It would also be wrong to call produce_points_to_chunk_ with a points_to_kind = MaybeUninit because then the value should be an `option<S>` (i.e. either `none` or fully initialized) , not a ctor_opt value.) The easy fix seems to be to continue calling produce_points_to_chunk_ with value `None` (that's OCaml's None) in this case for now.

src/verifast.ml
```diff
@@ -657,6 +657,43 @@ module VerifyProgram(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
       with_context (Executing (h, env, l, "Consuming object")) $. fun () ->
       consume_c_object_core_core l real_unit_pat pointerTerm pointeeType h env true false $. fun _ h (Some value) ->
       cont (Chunk ((generic_points_to_symb (), true), [pointeeType], real_unit, [pointerTerm; value], None)::h) env
+    | ExprStmt (CallExpr (l, "open_malloc_block", targs, [], args, Static)) when language = CLang ->
+      require_pure ();
+      let e = match (targs, args) with ([], [LitPat e]) -> e | _ -> static_error l "open_malloc_block expects no type arguments and one argument." None in
+      let (w, tp) = check_expr (pn,ilist) tparams tenv e in
+      let sn, targs = match tp with PtrType (StructType (sn, targs)) -> sn, targs | _ -> static_error l "The argument of open_malloc_block must be of type pointer-to-struct." None in
+      let _, _, body_opt, Some padding_pred_symb, _ = List.assoc sn structmap in
```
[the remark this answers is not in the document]
NielsMommen: I changed it now to not allow to use `open_malloc_block(s)` when no padding symbol is found for `s`, which also includes packed structs. Is this fine?

src/verify_expr.ml
NielsMommen: I changed the `assume_bounds` behaviour as suggested and the option field type is no longer passed.

btj: It seems as if you *are* still passing the option field type?

src/verify_expr.ml
```diff
@@ -1841,30 +1858,33 @@ module VerifyExpr(VerifyProgramArgs: VERIFY_PROGRAM_ARGS) = struct
...
+          cont chunks h (Some struct_val)
         | (f, (lf, gh, t0, offset, finit))::fields ->
           let t = instantiate_type tpenv t0 in
           match t with
             StaticArrayType (_, _) | StructType _ | UnionType _ ->
             consume_c_object_core_core l coefpat (field_address l typeid_env addr sn targs f) t h typeid_env true consumeUninitChunk $. fun chunks' h (Some value) ->
-            let value =
-              if consumeUninitChunk then value else prover_convert_term value t t0
-            in
+            let value = prover_convert_term value t t0 in
```
btj, line 1870: If `consumeUninitChunk` is true, `value` is not of type `t`, right? So in that case it's not okay to convert from t to t0?

src/verify_expr.ml
NielsMommen: Forgot to push that change, my apologies!

btj (approved):

PR #750, opened 2025-03-10
refinement-checker: Move out of box -> Box::into_inner

btj:


btj:
The code for `Box::into_inner` used by the refinement checker is wrong if T has a destructor.

PR #755, opened 2025-03-14
Add a useful new lemma_auto in list.gh

etiennebirling:
trivially, for L a list, if you replace L's i-th element by L's i-th element, the result is L


btj:
Thanks for your contribution! Note: a proof of the lemma in `list.c` is missing.

etiennebirling:
updated

PR #768, opened 2025-04-14
lemma_auto list_map_equal_length

etiennebirling:


btj:
I now realize that this whole `-check_vfmanifest` business is from an age when VeriFast did not yet ship with Z3 and users had to (optionally) install it separately. This no longer applies; I finally removed this nasty hurdle to updating `list.c`. Should have done this much sooner!

etiennebirling:
So one is not required anymore to write a proof in list.c? Or was it rather this bitops thing that you solved?

btj:
The bitops thing.

PR #771, opened 2025-04-21
Ref_cell.rs: Immutable borrows

Mmaarten23:


btj:
Please add the example to `tests/rust/testsuite.mysh`.

Mmaarten23:
The example was already added to the testsuite in the previous version with just mutable borrows: https://github.com/verifast/verifast/commit/a3fbf572f457ef163676403d2907669be65416e2#diff-5271c3332a7297f2a424ecd4d3a917dc996a6ee5cdf622ea652ab9fb1e519447

I don't see anything special that would need to be added to make this version work?

btj:
> I don't see anything special that would need to be added to make this version work?

You're right; my bad! Looks good! Congrats!!

PR #998, opened 2026-03-01
Add documentation for non-existing Rust functions

jake-87:
I hope this is some suitable documentation? Please feel free to edit anything you think could be worded better.


btj:
No, this is not the right place to put this information. I suggest you add a separate page "Error messages" to the reference. Note also that there is no point in discussing IDE versus command line, they are exactly the same error message!

PR #1001, opened 2026-03-15
[Rust] Add support for &[T] shared slice reference types

jrey8343:
## Summary

Adds ownership of shared slice references, `&[T]`, to the Rust front end: `own` of an `&[T]` value is `[_]<[T]>.share(lft, tid, v)`, the same shape as `&str`'s `own` uses `<str>.share`, and `<[T]>.share` is defined in the prelude per element via `<T>.share`. Related to #997 (that issue is about `&mut [u8]`, which is not covered here; see below).

**Motivation:** the [verify-rust-std](https://github.com/model-checking/verify-rust-std) challenges on `slice`, slice iterators and `Vec` (17, 18, 23, 24) all stop at `Expressing ownership of &[_] values is not yet supported` on the first `&[T]` parameter; the upstream `vec` proof in `tests/rust/safe_abstraction/vec/verified` avoids every slice-returning function for the same reason.

### `bin/rust/prelude_core.rsspec`
- `pred array_share<T>(k, t, l: *T, length)`: per-element shared ownership, `[_]<T>.share(k, t, l)` for each element. This is the predicate from the `vec` proof, body for body, moved into the prelude.
- `pred_ctor slice_share<T>()(k, t, l: *[T]) = array_share::<T>(k, t, l as *T, l.len())` and `type_pred_def for<T> <[T]>.share = slice_share::<T>`.

### `src/rust_frontend/vf_mir_translator/vf_mir_translator.ml`
- `slice_ref_ty_info`: for `&[T]`, `own` produces `[_]<[T]>.share(lft, tid, v)` instead of `Error`. `&mut [T]` still reports "not yet supported".

### `tests/rust/safe_abstraction/vec/verified/vec/mod.rs`
- The local `array_share` definition is removed (it now comes from the prelude); `array_share_mono` and `array_sync` stay and resolve against it.

### `tests/rust/safe_abstraction/slice_ref.rs`, `tests/rust/testsuite.mysh`
- Test of `&[u8]` pass-through with a `<[u8]>.share` spec, registered in the suite.

## Not in this PR

`&mut [T]`: its ownership would go through `<[T]>.full_borrow_content`, which is not defined (an earlier revision of this PR referenced it anyway and failed the prototype implementation check once the test ran). The `array_at_lft_fbc` predicate in the `vec` proof looks like the right basis for it; happy to do that as a follow-up.

## Verification

Fork CI on the rebased branch: the Rust test suite passes on Ubuntu 22/24, Windows and macOS arm64 (`slice_ref.rs` and the full `vec` proof, including both refinement-checker steps). The macOS Intel job fails at the `vfide` `/usr/local` library check, as it does on master.


jrey8343:
@btj, reviving this after the 26.09 release. The branch is now rebased onto current master (8a4c11f, on top of #1012's nightly-2026-02-05 bump and your #1008), and trimmed to the slice-reference change only: the const-generic, fn-as-operand, SwitchInt and closure commits that had accumulated here are out, since they have their own PRs (#1004, #1005, #1006, #1007). Net diff is now 50 lines over five files.

What it does now:
- `&[T]`: `own` is `[_]<[T]>.share(lft, tid, v)`, the same shape as `&str`'s `own` uses `<str>.share`.
- `<[T]>.share` is defined in the prelude per element via `<T>.share`, as you asked (no fractured borrow), so it is correct for element types with interior mutability. The underlying `array_share<T>` predicate is the one from your `vec` proof, body for body; since the prelude now defines it, the copy in `tests/rust/safe_abstraction/vec/verified/vec/mod.rs` is removed (the `array_share_mono` and `array_sync` lemmas stay there and resolve against the prelude).
- `&mut [T]` is **out** of this PR. The earlier revision routed its ownership through `<[T]>.full_borrow_content`, which is not defined, and the prototype implementation check of a safe fn taking `&mut [u8]` failed on it once the test actually ran. The translator reports `&mut [_]` as not yet supported, as master does. Your `array_at_lft_fbc` in the vec proof looks like the right basis for a follow-up.

Two fixes to the PR itself: `tests/rust/safe_abstraction/slice_ref.rs` was never listed in `tests/rust/testsuite.mysh`, so the suite had not been running it (it is registered now), and the test is reduced to the `&[T]` case.

CI on my fork for the pushed branch: the Rust test suite passes on Ubuntu 22/24, Windows and macOS arm64, including `slice_ref.rs` and the whole `vec` proof with both refinement-checker steps. The macOS Intel job fails at the `vfide` `/usr/local` library check (`! grep /usr/local vfide-libraries.log`), which master's own last two runs fail identically, so that one is the runner image, not the change. The PR title and description are rewritten for the reduced scope.

Why now: feliperodri asked on model-checking/verify-rust-std#651 whether a VeriFast route is viable for the slice-iterator and Vec challenges, and this is the missing piece. In your own `tests/rust/safe_abstraction/vec/verified` tree the six specified functions avoid slice references, while `as_slice`, `as_mut_slice`, `deref`, `deref_mut`, `extend_from_slice`, `leak` and `as_ref` all need `&[T]`/`&mut [T]` ownership to be expressible; on the verify-rust-std side, Challenge 17/18 (slice, slice iterators) and 23/24 (Vec) all stop at master's `Expressing ownership of &[_] values is not yet supported` on the first `&[T]` parameter. The sibling PRs are rebased the same way (#1004 on top of #1008's `Z` representation, with signed `Int` sign-extended; #1005 and #1006 apply cleanly) and I will update each of them once this one has your direction.

jrey8343:
Thanks for merging this and #1004/#1006. One question about timing: the verify-rust-std VeriFast harness (`verifast-proofs/setup-verifast-home`) only installs attested release tarballs, so the Vec and slice-iterator proofs there can leave draft only once a release contains `5e5fc0f8`. Is a release planned in the near term? If it would help, I can get the `&mut [T]` follow-up in first so a single release covers all of it.

btj:
Yes, I plan to do another Rust toolchain bump and associated release soon, to fix [this breakage](https://github.com/model-checking/verify-rust-std/actions/runs/34762158269).

PR #1004, opened 2026-03-19
[Rust] Support const generic Bool and Int types

jrey8343:
## Summary
- Extend `translate_generic_arg` to handle `Bool` and signed `Int` const generic arguments (#1008 already generalised the unsigned case to every width via `Z.of_uint128`).
- Signed values are sign-extended from their byte size with `Stdint.Int128` before `Z.of_int128`, so a negative argument is not read as a large unsigned one.
- Enables functions with const generic parameters like `const DELETED: bool` (e.g. `Vec::retain_mut`'s inner `process_loop`).

Split from #1002 per review feedback.

## Test
- `test_const_generic_bool.rs` — verifies `process::<false>(42)` and `process::<true>(10)` with postcondition `result == x`; registered in `tests/rust/testsuite.mysh`.

### DIFF
--- a/src/rust_frontend/vf_mir_translator/vf_mir_translator.ml
+++ b/src/rust_frontend/vf_mir_translator/vf_mir_translator.ml
@@ -1254,7 +1254,13 @@ module Make (Args : VF_MIR_TRANSLATOR_ARGS) = struct
               begin match ty, size with
                 {kind=UInt USize}, 8 ->
                   Ast.LiteralConstTypeExpr (loc, Stdint.Uint128.to_int v)
-              | _ -> 
+              | {kind=Bool}, 1 ->
+                  Ast.LiteralConstTypeExpr (loc, Stdint.Uint128.to_int v)
+              | {kind=UInt _}, _ ->
+                  Ast.LiteralConstTypeExpr (loc, Stdint.Uint128.to_int v)
+              | {kind=Int _}, _ ->
+                  Ast.LiteralConstTypeExpr (loc, Stdint.Uint128.to_int v)
+              | _ ->
                   failwith "Unsupported constant type or size"
               end
           | _ -> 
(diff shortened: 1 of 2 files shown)


src/rust_frontend/vf_mir_translator/vf_mir_translator.ml
```diff
@@ -1254,7 +1254,13 @@ module Make (Args : VF_MIR_TRANSLATOR_ARGS) = struct
               begin match ty, size with
                 {kind=UInt USize}, 8 ->
                   Ast.LiteralConstTypeExpr (loc, Stdint.Uint128.to_int v)
```
btj, line 1256: I notice a bug in the original code: not all values of Rust type `usize` (64 bits on a 64-bit platform) fit into an OCaml `int` (63-bit signed integer on a 64-bit platform). Please first submit a separate PR that fixes this bug. A concern is that `type_` values should be comparable using structural equality (`=`). `Big_int` values are not. The best fix would be to use the `Zarith` package, but adding dependencies to VeriFast is painful. So I propose to replace the `int` parameter of `LiteralConstTypeExpr` and `LiteralConstType` by a `string` that stores a base-10 textual representation of the integer value.

btj (changes requested):

src/rust_frontend/vf_mir_translator/vf_mir_translator.ml
btj: Update: I am fixing this bug myself: https://github.com/verifast/verifast/pull/1008

jrey8343: Thanks for that and good catch 👍

jrey8343:
Rebased onto master with your #1008 in. Since that already handles every unsigned width via `Z.of_uint128`, what remains here is the `Bool` arm and the signed `Int` arm; signed values are sign-extended from their byte size with `Stdint.Int128` before `Z.of_int128`, so a negative const generic argument is no longer read as a large unsigned one. The test file was never listed in `tests/rust/testsuite.mysh`; it is registered now and passes (fork CI green on Ubuntu 22/24, Windows and macOS arm64; the macOS Intel job fails at the `vfide` `/usr/local` check as on master).

btj (approved):

PR #1005, opened 2026-03-19
[Rust] Handle function items as operands in rvalues

jrey8343:
## Summary
- Function items (e.g., passing `Err` as a value in `.map_or(Ok(()), Err)`) are zero-sized types whose dispatch is handled by MIR monomorphization
- Produce a placeholder `IntLit(0)` instead of failing when a function item appears as an operand

Split from #1002 per review feedback.

## Test
- `test_fn_as_operand.rs` — passes `identity` as a `fn(i32) -> i32` argument


jrey8343:
Rebased onto master (26.09). The change applies cleanly; the test file was never listed in `tests/rust/testsuite.mysh`, so it is registered now, and its `main` no longer assumes `false`, so the call that passes `identity` as an operand is actually verified (the indirect call inside `apply` stays assumed away with `-allow_assume -allow_dead_code`, since calling through a fn pointer is not what this tests). Fork CI is green on Ubuntu 22/24, Windows and macOS arm64; the macOS Intel job fails at the `vfide` `/usr/local` check as it does on master.

tests/rust/safe_abstraction/test_fn_as_operand.rs
```diff
@@ -0,0 +1,23 @@
+// Test for functions as operands in rvalues
+// Previously failed with "Todo: Functions as operand in rvalues are not supported yet"
```
btj, line 2: Not true: this example goes through just fine without the change?

btj (changes requested):
Change not motivated: the test case goes through even without the change.

jrey8343:
You are right, and the test was the wrong one: `apply(identity, 42)` with a `fn(i32) -> i32` parameter is lowered through a `ReifyFnPointer` cast, which `translate_rvalue`'s Cast arm already handles, so the changed arm was never reached.

The arm *is* reached (master fails with `Todo: Functions as operand in rvalues are not supported yet`) when a fn item is used without coercion: passed to a generic parameter (`apply::<F: Copy>(identity, 42)`), stored in an aggregate (`Some(identity)`), or `r.err().map_or(Ok(()), Err)`. I tried three representations for the operand on my fork, with those three shapes as tests:

| operand value | generic param | aggregate | `map_or(.., Err)` |
|---|---|---|---|
| `0` (this PR) | type mismatch: `c_int` vs `identity` | same | passes |
| `identity` (the function name) | type mismatch: `*_` vs `identity` | same | passes |
| `identity` cast to `FuncType identity` | same `*_` vs `identity` | same | passes |

So the placeholder is not a fix: `translate_fn_def_ty` gives fn-item types the type `FuncType <name>`, and none of these values has that type. (Two side findings while testing: an `F: Fn(..)` bound crashes with `Pure function core::ops::FnOnce::Output_typeid missing in the runtime library`, and `let f = identity;` fails in the `Use` arm with `Invalid operand translation for Rvalue`.)

Unless you have a preferred representation for fn-item values that I should implement, I will close this PR. It is not on the path of the Vec/slice work, which now depends on #1015.

jrey8343:
Closing per the findings above; happy to reopen or redo it against a representation you prefer.

PR #1006, opened 2026-03-19
[Rust] Translate Assert terminators

jrey8343:
## Summary

Translate rustc's `Assert` terminator. Since #651 the exporter accepts it but exports it without its data (`assert @9: Void`), and the translator has no arm for it, so any body with a non-constant division (the divide-by-zero check, plus the overflow check for signed division) fails with `Pattern matching failed` in `translate_terminator_kind`. Indexing with a non-constant index lowers to the same terminator.

## Changes

- `vf_mir.capnp`: `AssertData { cond: Operand, expected: Bool, target: BasicBlockId, unwindAction: UnwindAction }`; the `assert` union member carries it.
- Exporter: encodes the four fields (`msg` is dropped).
- Translator: `if cond == expected { goto target } else { <unwind action> }`; under `-ignore_unwind_paths` the panicking branch is an abort, as the `Terminate` unwind action already is. The check thus becomes a path condition for the operation it guards, so a safe `x / y` verifies with `req true`.
- Refinement checker: prints the terminator and compares the two sides (condition operand, expected value, target; the unwind path is not followed, as for `Drop`).
- Test: `test_division.rs` (`x / 2` and unsigned `x / y`, safe fns with `req true` and `ens result == x / y`), registered in `tests/rust/testsuite.mysh`. Signed division is not covered: rustc lowers its overflow check to a `BitAnd` of two booleans, which `translate_bin_op` maps to the integer operator ("Expression of arithmetic or pointer type expected"); that needs operand types in the binary-operation translation and is a separate change.

The earlier version of this PR unwrapped `ConstTypeExpr` in `translate_sw_int` for the same crash at a different site; that path is no longer reachable on master since the exporter change, and the unwrapping plus a `static_error` catch-all are kept only as defensive code.

## Verification

Fork CI on the branch: Rust test suite green on Ubuntu 22/24, Windows and macOS arm64; the macOS Intel job fails at the `vfide` `/usr/local` check, as on master.


jrey8343:
Rebased onto master and reworked, because the crash moved. On master the `SwitchInt` path this PR originally patched is no longer reachable: since #651 the exporter emits rustc's `Assert` terminator as a payload-less `assert @9: Void`, and the translator has no arm for it, so `x / y` (or `a[i]`) with a non-constant operand fails with `Pattern matching failed` in `translate_terminator_kind`. The branch now adds `AssertData` (condition, expected value, target, unwind action) to the schema, exports it, translates it as `if cond == expected { goto target } else { unwind }` (an abort under `-ignore_unwind_paths`, as for the `Terminate` action), and teaches the refinement checker to print and compare it. The check becomes a path condition for the guarded operation, so a safe `fn div(x: usize, y: usize) -> usize { x / y }` verifies with `req true; ens result == x / y;`. Signed division is still out: its overflow check is a `BitAnd` of two booleans, which the binary-operation translation maps to the integer operator; that needs operand types there and I have left it for a separate change. Title and description updated; fork CI green on the four working runners (macOS Intel fails the `vfide` `/usr/local` check as on master). If you would rather model the panicking branch differently under `-ignore_unwind_paths`, that is a one-line change in the translator arm.

